PR #15, opened 2021-11-09
Created Scooter model

### LINKED ISSUE #1: Create Scooter Model (M)
Scooter : 
- serialNumber: String
- currentLocation: { latitude: Double, longitude: Double }
- batteryLevel: { level: Double, BatteryStatus(ENUM): LOW=[0, 30), MEDIUM=[30,60), HIGH=[60, 100] }
- status - AVAILABLE, RESERVED

dbernatt:
Implemented the Scooter model.
Scooter
- serialNumber (String)
- currentLocation (Location: latitude(Double), longitude(Double))
- battery (Battery: level(Double), batteryStatus(BatteryStatus Enum: LOW, MEDIUM, HIGH))
- status (ScooterStatus Enum: AVAILABLE, RESERVED)


nymphadora-tonks-team (commented):
We would like not to use lombok. You can find some details regarding the reasons for this in https://medium.com/@vgonzalo/dont-use-lombok-672418daa819

birleadrian:
I see that id attribute is String. Shouldn't it be long type?

dbernatt:
> I see that id attribute is String. Shouldn't it be long type?

Depends what type we will use: UUID or long id. But I don't know yet.
Also we can add a BaseEntity for each model which contains the id, so we don't need to include all the time the id. 
Every model will extend this BaseEntity which adds an id for each model. I recommend to use UUID type for id, because it has a big advantage over autoincremented ids. You can read about this [here](https://mareks-082.medium.com/auto-increment-keys-vs-uuid-a74d81f7476a).

dbernatt:
> > I see that id attribute is String. Shouldn't it be long type?
> 
> Depends what type we will use: UUID or long id. But I don't know yet. Also we can add a BaseEntity for each model which contains the id, so we don't need to include all the time the id. Every model will extend this BaseEntity which adds an id for each model. I recommend to use UUID type for id, because it has a big advantage over autoincremented ids. You can read about this [here](https://mareks-082.medium.com/auto-increment-keys-vs-uuid-a74d81f7476a).

Also we can create Abstract Model which stores the hash() and equals() methods. So only the relevant information remains in each Model. But I don't know how firebase works with these classes yet. Just a suggestion,

src/main/java/com/bootcamp/demo/model/Scooter.java
```diff
@@ -0,0 +1,74 @@
...
+        sb.append("serialNumber='").append(serialNumber).append(", ");
+        sb.append("currentLocation=").append(currentLocation).append(", ");
+        sb.append("battery=").append(battery).append(", ");
+        sb.append("status=").append(status).append('}');
+        return sb.toString();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
nymphadora-tonks-team: I know that this method was generated as such, but it is always good to check it and see if there is anything that you would like to change.
More specifically here, what could be an issue with the fact that `if` is used without braces? Applies for other occurrences as well.

src/main/java/com/bootcamp/demo/model/Scooter.java
```diff
@@ -0,0 +1,74 @@
...
+        if (o == null || this.getClass() != o.getClass()) return false;
+        final var scooter = (Scooter) o;
+        return this.getSerialNumber().equals(scooter.getSerialNumber())
```
nymphadora-tonks-team: Since for the moment (at least) there is no validation for the values of the fields, what issue could occur here? Applies for other occurrences as well.

src/main/java/com/bootcamp/demo/model/util/Battery.java
```diff
@@ -0,0 +1,48 @@
+package com.bootcamp.demo.model.util;
+
+import java.util.Objects;
+
+public final class Battery {
+    private Double level;
+    private BatteryStatus status;
```
nymphadora-tonks-team: Please highlight why you have decided to save this information in 2 separate fields. Notice that it would be very easy to construct an invalid instance, e.g. with level = 100 (since you mentioned that 100 would be the highest value) & status = LOW.

src/main/java/com/bootcamp/demo/model/util/BatteryStatus.java
```diff
@@ -0,0 +1,17 @@
+package com.bootcamp.demo.model.util;
+
+public enum BatteryStatus {
+    LOW("LOW"),
```
nymphadora-tonks-team: Please highlight why you have decided to set for value the same as the name.

src/main/java/com/bootcamp/demo/model/Scooter.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.Objects;
+
+public final class Scooter {
+    private String serialNumber;
+    private Location currentLocation;
+    private Battery battery;
+    private ScooterStatus status;
+
+    private Scooter() {}
```
nymphadora-tonks-team: Please highlight why have made this private and if you think that it might cause any issues.

src/main/java/com/bootcamp/demo/model/util/Battery.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public Battery() {}
```
nymphadora-tonks-team: Highlight why you have added a default constructor.

src/main/java/com/bootcamp/demo/model/util/Location.java
```diff
@@ -0,0 +1,48 @@
+package com.bootcamp.demo.model.util;
+
+import java.util.Objects;
+
+public final class Location {
+    private Double latitude;
+    private Double longitude;
+
+    public Location() {};
```
nymphadora-tonks-team: Why have you added `;` after `}`?

src/main/java/com/bootcamp/demo/model/util/Battery.java
```diff
@@ -0,0 +1,48 @@
+package com.bootcamp.demo.model.util;
```
nymphadora-tonks-team: Highlight why you have added these classes under `util` package.

src/main/java/com/bootcamp/demo/model/Scooter.java
dbernatt: Ups. My bad. If it has private then other classes cannot create instances of this class.

src/main/java/com/bootcamp/demo/model/Scooter.java
dbernatt: Maintainability-wise, it's always smarter to use braces to prevent any mistakes(ex. after this if statement we add another statement which always executes if the earlier check was false). In this example it works but, its just a good practice I guess to use braces all the time.

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: In my opinion this is cleaner code. We break up the base classes into pieces. Because its more structured this way and we can see right away that what base models we have (ex. Scooter, User, Booking). Make sens no?

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: The validation of the fields its not implemented yet. But it will be implemented on the service layer.

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: https://stackoverflow.com/questions/50114944/why-we-need-an-empty-constructor-to-passing-save-a-data-from-firebase
But its my bad bc, in our case we don't have constructor with args. So I'll remove it.

src/main/java/com/bootcamp/demo/model/Scooter.java
dbernatt: But I remove the noarg constructor bc didn't make sens if we don't have any other constructor.

src/main/java/com/bootcamp/demo/model/Scooter.java
dbernatt: Back in 2014 was a famous Apple bug that was caused by this mistake. https://www.imperialviolet.org/2014/02/22/applebug.html

src/main/java/com/bootcamp/demo/model/Scooter.java
dbernatt: We need to check if the serialNumber is not null, right? In the constructor we need to add a non null check?
Before calling equals() we need to check if the objects are not null.

src/main/java/com/bootcamp/demo/model/util/BatteryStatus.java
dbernatt: http://www.vineetmanohar.com/2010/01/3-ways-to-serialize-java-enums/
For serialization and deserialization purposes. ^

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: Include the validation in this PR?

dbernatt (commented):
Fixed some comments and answered questions.

src/main/java/com/bootcamp/demo/model/util/Battery.java
nymphadora-tonks-team: We can have sub-packages for the `model` package, but perhaps a more suitable name could be found? Using util usually suggests some utility classes.

src/main/java/com/bootcamp/demo/model/util/Battery.java
nymphadora-tonks-team: I was not thinking from a validation perspective, but rather from a modelling one. Why not have the level encapsulated in `BatteryStatus`?

src/main/java/com/bootcamp/demo/model/util/BatteryStatus.java
nymphadora-tonks-team: This comment was related with this one: https://github.com/nymphadora-tonks-team/booking-jb/pull/15/files#r747761222, as I was thinking if there could be a model that would accommodate both serialising enough details & being intuitive.

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: I changed to component. Its all good?

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: The BatteryStatus is an enum with a fixed string category name("LOW", "MEDIUM", "HIGH"). If I add the level (Double) attribute, then, for ex. I want to change the category name because the level is not compatible with the present one, then  I can't do that because its fixed. 

Or, I don't understand how do you want to do that. Can you explain it more?

> Enum constants are implicitly static and final and you can not change their value once created. Enum in Java provides type-safety and can be used inside switch statements like int variables. 

I tried this: 
` public enum BatteryStatus {
    LOW("LOW", 0.0),
    MEDIUM("MEDIUM", 30.0),
    HIG("HIGH", 60.0);

    private final String value;
    private Double level;

    BatteryStatus(String value, Double level) {
        this.value = value;
        this.level = level;
    }

    public Double getLevel() {
        return level;
    }

    public String getValue() { return value; }
}
`

src/main/java/com/bootcamp/demo/model/util/BatteryStatus.java
dbernatt: The Battery class model is right know include all the necessary information, and for me is intuitive :). Contains the level of the battery and the status which is categorized in 3 types.

src/main/java/com/bootcamp/demo/model/util/Battery.java
dbernatt: I can't change the status category inside an enum based on the level value. Thats why I separated these 2 values in the Battery class.

loredanaiamnitchi (approved):

vladmsn (approved):

mariatirlea (approved):

birleadrian (approved):

nymphadora-tonks-team (approved):

PR #16, opened 2021-11-09
Created Booking Model #2

### LINKED ISSUE #2: Create Booking Model (M)
Draft version:
Booking
- id
-serialNumber (scooter)
- accountId
- startDate
- endDate
- payment (SUCCESS/FAILURE)
- cost?

birleadrian:
I've created the Booking Model #2 using Lombok to generate Getters , Setters and Constructors.

### DIFF
--- a/src/main/java/com/bootcamp/demo/models/Booking.java
+++ b/src/main/java/com/bootcamp/demo/models/Booking.java
@@ -1,0 +1,23 @@
+package com.bootcamp.demo.models;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import java.util.Date;
+
+
+@Getter
+@Setter
+@AllArgsConstructor
+@NoArgsConstructor
+public class Booking {
+
+    private long id;
+    private long serialNumber;
+    private long accountId;
+    private Date startDate;
+    private Date endDate;
+    private PaymentStatus payment;
+
+}
--- a/src/main/java/com/bootcamp/demo/models/Booking.java
+++ b/src/main/java/com/bootcamp/demo/models/Booking.java
@@ -1,18 +1,9 @@
 package com.bootcamp.demo.models;
-
-import lombok.AllArgsConstructor;
-import lombok.Getter;
-import lombok.NoArgsConstructor;
-import lombok.Setter;
 import java.util.Date;
+import java.util.Objects;
 
 
-@Getter
-@Setter
-@AllArgsConstructor
-@NoArgsConstructor
 public class Booking {
-
     private long id;
     private long serialNumber;
     private long accountId;
@@ -20,4 +11,77 @@ public class Booking {
     private Date endDate;
     private PaymentStatus payment;
 
+    public Booking(){}
+    public Booking(long id,long serialNumber, long accountId,Date startDate,Date endDate,PaymentStatus payment){
+        this.id = id;
+        this.serialNumber = serialNumber;
+        this.accountId = accountId;
+        this.startDate = startDate;
+        this.endDate = endDate;
+        this.payment = payment;
+    }
+
+    //Getters
+    public long getId() {
+        return id;
+    }
+
+    public long getSerialNumber() {
+        return serialNumber;
+    }
+
+    public long getAccountId() {
+        return accountId;
+    }
+
+    public Date getStartDate() {
+        return startDate;
+    }
+
+    public Date getEndDate() {
+        return endDate;
+    }
+
+    public PaymentStatus getPayment() {
+        return payment;
+    }
+
+    //Setters
+
+    public void setId(long id) {
+        this.id = id;
+    }
+
+    public void setSerialNumber(long serialNumber) {
+        this.serialNumber = serialNumber;
+    }
+
+    public void setAccountId(long accountId) {
+        this.accountId = accountId;
+    }
+
+    public void setStartDate(Date startDate) {
+        this.startDate = startDate;
+    }
+
+    public void setEndDate(Date endDate) {
+        this.endDate = endDate;
+    }
+
+    public void setPayment(PaymentStatus payment) {
+        this.payment = payment;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Booking booking = (Booking) o;
+        return id == booking.id && serialNumber == booking.serialNumber && accountId == booking.accountId && startDate.equals(booking.startDate) && endDate.equals(booking.endDate) && payment == booking.payment;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, serialNumber, accountId, startDate, endDate, payment);
+    }
 }
--- a/src/main/java/com/bootcamp/demo/models/Booking.java
+++ b/src/main/java/com/bootcamp/demo/models/Booking.java
@@ -1,18 +1,20 @@
 package com.bootcamp.demo.models;
-import java.util.Date;
+import java.time.LocalDateTime;
 import java.util.Objects;
+import java.util.UUID;
 
 
-public class Booking {
-    private long id;
-    private long serialNumber;
-    private long accountId;
-    private Date startDate;
-    private Date endDate;
+public final class Booking {
+    private UUID id;
+    private UUID serialNumber;
+    private UUID accountId;
+    private LocalDateTime startDate;
+    private LocalDateTime endDate;
     private PaymentStatus payment;
 
-    public Booking(){}
-    public Booking(long id,long serialNumber, long accountId,Date startDate,Date endDate,PaymentStatus payment){
+    public Booking() {}
+
+    public Booking(UUID id,UUID serialNumber, UUID accountId,LocalDateTime startDate,LocalDateTime endDate,PaymentStatus payment) {
         this.id = id;
         this.serialNumber = serialNumber;
         this.accountId = accountId;
@@ -21,24 +23,23 @@ public class Booking {
         this.payment = payment;
     }
 
-    //Getters
-    public long getId() {
+    public UUID getId() {
         return id;
     }
 
-    public long getSerialNumber() {
+    public UUID getSerialNumber() {
         return serialNumber;
     }
 
-    public long getAccountId() {
+    public UUID getAccountId() {
         return accountId;
     }
 
-    public Date getStartDate() {
+    public LocalDateTime getStartDate() {
         return startDate;
     }
 
-    public Date getEndDate() {
+    public LocalDateTime getEndDate() {
         return endDate;
     }
 
@@ -46,25 +47,23 @@ public class Booking {
         return payment;
     }
 
-    //Setters
-
-    public void setId(long id) {
+    public void setId(UUID id) {
         this.id = id;
     }
 
-    public void setSerialNumber(long serialNumber) {
+    public void setSerialNumber(UUID serialNumber) {
         this.serialNumber = serialNumber;
     }
 
-    public void setAccountId(long accountId) {
+    public void setAccountId(UUID accountId) {
         this.accountId = accountId;
     }
 
-    public void setStartDate(Date startDate) {
+    public void setStartDate(LocalDateTime startDate) {
         this.startDate = startDate;
     }
 
-    public void setEndDate(Date endDate) {
+    public void setEndDate(LocalDateTime endDate) {
         this.endDate = endDate;
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import java.util.Date;
+
+
+@Getter
+@Setter
+@AllArgsConstructor
+@NoArgsConstructor
```
dbernatt: You can use `@Data` instead of `@Getter`,` @Setter`. 
You can read about `@Data` annotation here -> [lombok @data](https://projectlombok.org/features/Data)

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,23 @@
...
+public class Booking {
+
+    private long id;
+    private long serialNumber;
+    private long accountId;
```
dbernatt: I think account id should be an UUID type.

src/main/java/com/bootcamp/demo/models/PaymentStatus.java
```diff
@@ -0,0 +1,6 @@
+package com.bootcamp.demo.models;
+
+public enum PaymentStatus {
+    SUCCESS,
+    FAILURE
```
dbernatt: [java enum serialization and deserialization](http://www.vineetmanohar.com/2010/01/3-ways-to-serialize-java-enums/)
You need to add a constructor where you specify the value of each enum category. You can read about this on the link above :).

> What does serialization and deserialization of enum mean? Serialization involves converting the enum to a value (usually primitive or a String) which can be easily stored on the disk or a database. Deserialization involves reading a stored value and converting it back to an enum.

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
```diff
@@ -29,7 +29,7 @@ public void configure(final HttpSecurity http) throws Exception {
     @Override
     protected void configure(final AuthenticationManagerBuilder auth) throws Exception {
         final var inMemoryAuth = auth.inMemoryAuthentication();
-        inMemoryAuth.withUser("test.team@gmail.com").password("{noop}xxx").roles("USER");
+        inMemoryAuth.withUser("").password("{noop}xxx").roles("USER");
```
dbernatt: The username and password needs to be changed?

nymphadora-tonks-team (commented):
We would like not to use lombok. You can find some details regarding the reasons for this in https://medium.com/@vgonzalo/dont-use-lombok-672418daa819

birleadrian:
I think for ID we can use an UUID cause it's immutable.

dbernatt (changes requested):
I reviewed your code. Please check out my comments.

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
+package com.bootcamp.demo.models;
+import java.util.Date;
+import java.util.Objects;
+
+
+public class Booking {
```
dbernatt: Class should be final

dbernatt (changes requested):

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
...
+    private long id;
+    private long serialNumber;
```
dbernatt: serialNumber should be String type

dbernatt (changes requested):

src/main/java/com/bootcamp/demo/models/Booking.java
birleadrian: We were specified not to use lombok

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
birleadrian: I didn't create another user , maybe IntelliJ or something

src/main/java/com/bootcamp/demo/models/Booking.java
birleadrian: serialNumber should be a number because we will map into database

src/main/java/com/bootcamp/demo/models/Booking.java
birleadrian: I already said that we can use UUID for id , but I'm waiting for another opinion

src/main/java/com/bootcamp/demo/models/Booking.java
nymphadora-tonks-team: Highlight why you think it is necessary to have both id and serial number.

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
...
+    private long accountId;
+    private Date startDate;
+    private Date endDate;
+    private PaymentStatus payment;
+
+    public Booking(){}
+    public Booking(long id,long serialNumber, long accountId,Date startDate,Date endDate,PaymentStatus payment){
```
nymphadora-tonks-team: Try to ensure that the class is formatted. here for example having an empty line between the constructors and a space between `)` and `{`. That way, it will be easier improve readability.
Note that there is a keyboard shortcut which does this automatically. You can find all shortcuts here: https://resources.jetbrains.com/storage/products/intellij-idea/docs/IntelliJIDEA_ReferenceCard.pdf

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
+package com.bootcamp.demo.models;
+import java.util.Date;
+import java.util.Objects;
+
+
+public class Booking {
+    private long id;
+    private long serialNumber;
+    private long accountId;
+    private Date startDate;
```
nymphadora-tonks-team: Generally, it is best to avoid using this class, especially when there are other alternatives. You can find details regarding the reasons why it is best not to use it & details about a new API available starting with java 8 here: https://programminghints.com/2017/05/still-using-java-util-date-dont/, and further details about java 8 api here: https://www.baeldung.com/migrating-to-java-8-date-time-api.

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
...
+    public Booking(long id,long serialNumber, long accountId,Date startDate,Date endDate,PaymentStatus payment){
+        this.id = id;
+        this.serialNumber = serialNumber;
+        this.accountId = accountId;
+        this.startDate = startDate;
+        this.endDate = endDate;
+        this.payment = payment;
+    }
+
+    //Getters
```
nymphadora-tonks-team: Could you please highlight why you have added this comment?

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
...
+        this.endDate = endDate;
+    }
+
+    public void setPayment(PaymentStatus payment) {
+        this.payment = payment;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
nymphadora-tonks-team: I know that this method was generated as such, but it is always good to check it and see if there is anything that you would like to change.
More specifically here, what could be an issue with the fact that if is used without braces? Applies for other occurrences as well.

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        Booking booking = (Booking) o;
+        return id == booking.id && serialNumber == booking.serialNumber && accountId == booking.accountId && startDate.equals(booking.startDate) && endDate.equals(booking.endDate) && payment == booking.payment;
```
nymphadora-tonks-team: - Since for the moment (at least) there is no validation for the values of the fields, what issue could occur here? Applies for other occurrences as well.
- Try to avoid lines that are very long, as they tend to be hard to follow. Try to separate these on a separate lines in a way which would be easier to follow.

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
nymphadora-tonks-team: Would it be possible that instead of copy, you cut the value?

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,87 @@
+package com.bootcamp.demo.models;
```
nymphadora-tonks-team: Could you highlight why you have chosen plural for the package name?

birleadrian: I used the plural for name , because we will have more models.

src/main/java/com/bootcamp/demo/models/Booking.java
dbernatt: But if its number than you have problems if the number is too large and don't fit in your variable, Its more secure to use String bc you don't have this kind of problems. in my pov.

src/main/java/com/bootcamp/demo/models/Booking.java
dbernatt: Here is the explanation. [link](https://softwareengineering.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural)
_Use the plural for packages with homogeneous contents and the singular for packages with heterogeneous contents._

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,78 @@
...
+    public void setEndDate(LocalDateTime endDate) {
+        this.endDate = endDate;
+    }
+
+    public void setPayment(PaymentStatus payment) {
+        this.payment = payment;
+    }
+
+    @Override
+    public int hashCode() {
```
nymphadora-tonks-team: You could have left the implementation for equals, with small updates for it to ensure that it will not cause any potential issues.

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,78 @@
...
+    private UUID id;
+    private UUID serialNumber;
+    private UUID accountId;
+    private LocalDateTime startDate;
+    private LocalDateTime endDate;
+    private PaymentStatus payment;
+
+    public Booking() {}
+
+    public Booking(UUID id,UUID serialNumber, UUID accountId,LocalDateTime startDate,LocalDateTime endDate,PaymentStatus payment) {
```
nymphadora-tonks-team: There are still some formatting inconsistencies, e.g. missing spaces after some `,`. When multiple persons collaborate on a project, it is easier to have it formatted consistently by all collaborators (for example the default formatting for intellij or a custom one can be used), as it improves readability (long term, you get accustomed to a certain formatting) & reduces the chances of being formatted only later on, when changes are made to this class.
IntelliJ keymap: https://resources.jetbrains.com/storage/products/intellij-idea/docs/IntelliJIDEA_ReferenceCard.pdf - includes also the key shortcut for "Reformat code", which will do this automatically.

src/main/java/com/bootcamp/demo/models/Booking.java
nymphadora-tonks-team: Sorry for the confusion. Forgot that for booking both an id is needed to identify it & a serialNumber to identify the scooter that was used for that booking.

src/main/java/com/bootcamp/demo/models/Booking.java
dbernatt: The serial number is a String please. change it.

dbernatt (changes requested):

src/main/java/com/bootcamp/demo/models/Booking.java
```diff
@@ -0,0 +1,102 @@
...
+    private UUID serialNumber;
+    private UUID accountId;
+    private LocalDateTime startDate;
+    private LocalDateTime endDate;
+    private PaymentStatus payment;
+
+    public Booking() {
+    }
+
+    public Booking(UUID id, UUID serialNumber, UUID accountId, LocalDateTime startDate, LocalDateTime endDate, PaymentStatus payment) {
```
dbernatt: The booking id needs to be a random generated UUID,therefore you don't need to include to the constructor parameter list.

dbernatt (changes requested):

src/main/java/com/bootcamp/demo/models/Booking.java
nymphadora-tonks-team: We could go with using singular and result in having only 1 model package.

src/main/java/com/bootcamp/demo/models/Booking.java
nymphadora-tonks-team: Also, as pointed out by @mariatirlea on slack, we could have the fields/params which are now UUID as String, and rely on other classes to do the generation as appropriate (here assign a UUID's string value). What do you think?

nymphadora-tonks-team (approved):
Provided that all comments are addressed before merge.

PR #19, opened 2021-11-10
Created User Model

### LINKED ISSUE #3: Create User Model (M)
Draft version:
User
- accountId
- email
- firstName

vladmsn:
Implemented user model according to the draft version (https://github.com/nymphadora-tonks-team/booking-jb/issues/3#issue-1048864429)

### DIFF
--- a/src/main/java/com/bootcamp/demo/model/User.java
+++ b/src/main/java/com/bootcamp/demo/model/User.java
@@ -40,19 +40,19 @@ public final class User {
         return lastName;
     }
 
-    public User setUserId(@NotNull UUID id) {
+    public User setUserId(@NotNull final UUID id) {
         return new User(id, email, firstName, lastName);
     }
 
-    public void setEmail(@NotNull String email) {
+    public void setEmail(@NotNull final String email) {
         this.email = email;
     }
 
-    public void setFirstName(@NotNull String firstName) {
+    public void setFirstName(@NotNull final String firstName) {
         this.firstName = firstName;
     }
 
-    public void setLastName(@NotNull String lastName) {
+    public void setLastName(@NotNull final String lastName) {
         this.lastName = lastName;
     }
 
--- a/src/main/java/com/bootcamp/demo/model/User.java
+++ b/src/main/java/com/bootcamp/demo/model/User.java
@@ -12,7 +12,7 @@ public final class User {
     public User() {}
 
     public User(String email, String firstName, String lastName) {
-        this.userId = UUID.randomUUID().toString();
+        this.userId = UUID.randomUUID().toString().replace("-", "");
         this.email = Objects.requireNonNull(email);
         this.firstName = Objects.requireNonNull(firstName);
         this.lastName = Objects.requireNonNull(lastName);
--- a/src/main/java/com/bootcamp/demo/model/User.java
+++ b/src/main/java/com/bootcamp/demo/model/User.java
@@ -4,26 +4,28 @@ import java.util.Objects;
 import java.util.UUID;
 
 public final class User {
-    private final UUID userId;
+    private String userId;
     private String email;
     private String firstName;
     private String lastName;
 
+    public User() {}
+
     public User(String email, String firstName, String lastName) {
-        this.userId = UUID.randomUUID();
+        this.userId = UUID.randomUUID().toString();
         this.email = Objects.requireNonNull(email);
         this.firstName = Objects.requireNonNull(firstName);
         this.lastName = Objects.requireNonNull(lastName);
     }
 
-    public User(UUID userId, String email, String firstName, String lastName) {
+    public User(String userId, String email, String firstName, String lastName) {
         this.userId = userId;
         this.email = Objects.requireNonNull(email);
         this.firstName = Objects.requireNonNull(firstName);
         this.lastName = Objects.requireNonNull(lastName);
     }
 
-    public UUID getUserId() {
+    public String getUserId() {
         return userId;
     }
 
@@ -39,8 +41,8 @@ public final class User {
         return lastName;
     }
 
-    public User setUserId(final UUID id) {
-        return new User(id, email, firstName, lastName);
+    public void setUserId(final String userId) {
+        this.userId = userId;
     }
 
     public void setEmail(final String email) {
--- a/src/main/java/com/bootcamp/demo/model/User.java
+++ b/src/main/java/com/bootcamp/demo/model/User.java
@@ -12,7 +12,7 @@ public final class User {
     public User() {}
 
     public User(String email, String firstName, String lastName) {
-        this.userId = UUID.randomUUID().toString().replace("-", "");
+        this.userId = UUID.randomUUID().toString();
         this.email = Objects.requireNonNull(email);
         this.firstName = Objects.requireNonNull(firstName);
         this.lastName = Objects.requireNonNull(lastName);
--- a/src/main/java/com/bootcamp/demo/model/User.java
+++ b/src/main/java/com/bootcamp/demo/model/User.java
@@ -1,5 +1,6 @@
 package com.bootcamp.demo.model;
 
+import javax.validation.constraints.NotNull;
 import java.util.Objects;
 import java.util.UUID;
 
@@ -9,7 +10,14 @@ public final class User {
     private String firstName;
     private String lastName;
 
-    public User(UUID userId, String email, String firstName, String lastName) {
+    public User(@NotNull String email, @NotNull String firstName, @NotNull String lastName) {
+        this.userId = UUID.randomUUID();
+        this.email = email;
+        this.firstName = firstName;
+        this.lastName = lastName;
+    }
+
+    public User(@NotNull UUID userId, @NotNull String email, @NotNull String firstName, @NotNull String lastName) {
         this.userId = userId;
         this.email = email;
         this.firstName = firstName;
@@ -32,19 +40,19 @@ public final class User {
         return lastName;
     }
 
-    public User setUserId(UUID id) {
+    public User setUserId(@NotNull UUID id) {
         return new User(id, email, firstName, lastName);
     }
 
-    public void setEmail(String email) {
+    public void setEmail(@NotNull String email) {
         this.email = email;
     }
 
-    public void setFirstName(String firstName) {
+    public void setFirstName(@NotNull String firstName) {
         this.firstName = firstName;
     }
 
-    public void setLastName(String lastName) {
+    public void setLastName(@NotNull String lastName) {
         this.lastName = lastName;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
+package com.bootcamp.demo.models;
+
+import java.util.Objects;
+
+public final class User {
+    private final Long userId;
```
dbernatt: We need to decide how to store ids in our application, so this id could change.

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
...
+    private String email;
+    private String firstName;
+    private String lastName;
+
+    public User(Long userId, String email, String firstName, String lastName) {
+        this.userId = userId;
+        this.email = email;
```
dbernatt: The user information needs to be non null. For ex. the user email, firstname, lastname...etc. cant be null. Check the values here.

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
+package com.bootcamp.demo.models;
+
+import java.util.Objects;
+
+public final class User {
```
dbernatt: Why is this class final?

dbernatt: Checkout my #15 PR, where I created an AbstractModel class for the models.

dbernatt (changes requested):
Review done. Checkout my comments.

src/main/java/com/bootcamp/demo/models/User.java
dbernatt: The id should be generated automatically by firebase or we add a UUID random generator.

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
+package com.bootcamp.demo.models;
```
nymphadora-tonks-team: Could you highlight why you have chosen plural for the package name?

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public String getFirstName() {
+        return firstName;
+    }
+
+    public String getLastName() {
+        return lastName;
+    }
+
+    public User setUserId(Long id) {
```
nymphadora-tonks-team: Could you highlight why you have added instantiation logic in the setter?

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
...
+        this.firstName = firstName;
+    }
+
+    public void setLastName(String lastName) {
+        this.lastName = lastName;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
```
nymphadora-tonks-team: I know that this method was generated as such, but it is always good to check it and see if there is anything that you would like to change.
More specifically here, what could be an issue with the fact that if is used without braces? Applies for other occurrences as well.

src/main/java/com/bootcamp/demo/models/User.java
vladmsn: To be honest I haven't given it much thought at the time. I thought that it would contain more classes that are different types that are all models (homogeneous contents).

src/main/java/com/bootcamp/demo/models/User.java
vladmsn: Some of the issues I now know this use of `if` without braces could pose:
    1. if/else ambiguity - which I didn't consider to be a problem here as the return statement rules out the execution of code beyond that point.
    2. It could cause errors if later on if we decide to comment or add code on a certain `if` branch and don't add the braces.

src/main/java/com/bootcamp/demo/models/User.java
vladmsn: Noted. I'll make the change after we figure out what type of Id generation method we will use.

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
+package com.bootcamp.demo.models;
+
+import java.util.Objects;
+
+public final class User {
+    private final Long userId;
+    private String email;
```
loredanaiamnitchi: I think we will also need the location where the user is, to then look for the nearest scooters

loredanaiamnitchi (changes requested):

src/main/java/com/bootcamp/demo/models/User.java
dbernatt: Hmm. If we want to make a real-time experience yes, we need to know every time where is the user to quickly check the nearest scooters. If we don't want to do that, then we just need to use the location what is provided by the user when asks for nearest scooters. And we only updating the nearest scooters when the user asks explicitly again. What we will use then? We need to decide. :)

src/main/java/com/bootcamp/demo/models/User.java
dbernatt: Every model are different usually. So its not homogeneous package. :)

src/main/java/com/bootcamp/demo/models/User.java
dbernatt: Its a good practice which will prevents future errors if you use braces.

src/main/java/com/bootcamp/demo/models/User.java
nymphadora-tonks-team: @dbernatt  you could also include a URL to a more detailed explanation if you think it might a more detailed overview/explanation might help.

src/main/java/com/bootcamp/demo/model/User.java
```diff
@@ -0,0 +1,77 @@
+package com.bootcamp.demo.model;
+
+import java.util.Objects;
+import java.util.UUID;
+
+public final class User {
+    private final UUID userId;
+    private String email;
```
nymphadora-tonks-team: Wondering if with the current approach & having a small overview of firestore, if we should have the userId provided separately, or if we should infer it from the email. What do you think? We can also leave it as such, but I was wondering if you could also see some downsides which might need to be covered (potentially not for this app, but for one targeted at managing users) through validation?

src/main/java/com/bootcamp/demo/models/User.java
dbernatt: Here is the explanation : [link](https://github.com/nymphadora-tonks-team/booking-jb/pull/16#discussion_r748357408)

vladmsn: Thanks. I understand the difference now.

src/main/java/com/bootcamp/demo/model/User.java
vladmsn: I think the user class could have the `userId` as a field as in the current approach to be used for crud operations. The user data from the database could be assumed to be validated for our booking oriented purposes. Of course this is risky if the id is not valid. I can't quite grasp the 'big picture' downsides.
In my opinion, the validation of a new user instance should take place at Service/Controller level, maybe even in the My account part of the application when a user signs up.

nymphadora-tonks-team: I was thinking about perhaps resulting in creating 2 users with the same email (so validation would be needed for email). But indeed, as you mentioned it is not necessarily in scope for booking (since it will not be targeted at user management).
It was meant more as a note (perhaps it would be best to include that for future comments along these lines), for possible future consideration.

src/main/java/com/bootcamp/demo/model/User.java
```diff
@@ -0,0 +1,85 @@
+package com.bootcamp.demo.model;
+
+import javax.validation.constraints.NotNull;
```
nymphadora-tonks-team: Some notes regarding this:
https://stackoverflow.com/questions/34094039/using-notnull-annotation-in-method-argument (the first reply)
https://www.baeldung.com/javax-validation
https://www.baeldung.com/java-bean-validation-not-null-empty-blank

loredanaiamnitchi (approved):

src/main/java/com/bootcamp/demo/model/User.java
```diff
@@ -0,0 +1,85 @@
...
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+        final User other = (User) obj;
+        return Objects.equals(userId, other.userId);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(userId);
```
dbernatt: Please read about how to write a good hash method [here](https://www.sitepoint.com/how-to-implement-javas-hashcode-correctly/). You need some changes here.

src/main/java/com/bootcamp/demo/model/User.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Objects;
+import java.util.UUID;
+
+public final class User {
+    private final UUID userId;
+    private String email;
+    private String firstName;
+    private String lastName;
+
+    public User(@NotNull String email, @NotNull String firstName, @NotNull String lastName) {
```
dbernatt: Please highlight why you added the @NotNull annotation to the input paramteres?

dbernatt (changes requested):

dbernatt (approved):

dbernatt (approved):

src/main/java/com/bootcamp/demo/model/User.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.UUID;
+
+public final class User {
+    private final UUID userId;
+    private String email;
+    private String firstName;
+    private String lastName;
+
+    public User(String email, String firstName, String lastName) {
+        this.userId = UUID.randomUUID();
```
nymphadora-tonks-team: Same here as for booking. Should the type for `userId` be String to avoid the issue that @mariatirlea pointed out on slack?

nymphadora-tonks-team (approved):
Please see the comment related to whether `UUID` or `String` is more suitable for `userId` before merge.

src/main/java/com/bootcamp/demo/model/User.java
```diff
@@ -4,26 +4,28 @@
...
+    private String userId;
     private String email;
     private String firstName;
     private String lastName;
 
+    public User() {}
+
     public User(String email, String firstName, String lastName) {
-        this.userId = UUID.randomUUID();
+        this.userId = UUID.randomUUID().toString().replace("-", "");
```
dbernatt: Highlight why you need to replace the '-' in the userId(UUID)? Its not valid UUID then!

dbernatt (changes requested):

birleadrian (approved):

PR #20, opened 2021-11-11
Create, getAll, getById scooter

### LINKED ISSUE #17: Add scooter (M)
Expose a method to add a scooter.

### LINKED ISSUE #18: Retrieve existing scooters (M)
Expose a method to retrieve existing scooters.
- getScooterById(id) -> Scooter
- getScooters() -> Scooter[]

dbernatt:
Implemented functionality:

- [x] Retrieve all scooters from db 
- [x] Retrieve a scooter by serialNumber
- [x] Insert a new scooter


pom.xml
```diff
@@ -56,6 +56,12 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
nymphadora-tonks-team: As discussed on slack, please remove the usage of lombok. Ensure that related properties are removed as well.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.firebase.database.annotations.NotNull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.*;
+
+import javax.validation.Valid;
+
+@RestController
+@RequestMapping("/api/scooters")
+@Slf4j
```
nymphadora-tonks-team: Same here. Use logging differently, not through lombok. There were some resources sent on slack, but if you think further ones are needed, let us know.

nymphadora-tonks-team (commented):
Regarding the issues outlined in the PRs description
1) As mentioned also on slack, there is a cookie used by the application in order to identify the fact that the user is logged in. The name of this is `JSESSIONID` and you can find it in Application -> Cookies. When this cookie is not present, no calls can be made, as being logged in is a constraints for accessing the application.
2) This happens due to `Cross Site Request Forgery`. You can find more details about it here:  https://docs.spring.io/spring-security/site/docs/current/reference/html5/#csrf. This is caused by the fact that by default, swagger will not include the necessary header. There are 2 ways in which this can be solved:
     - Ensure that this header is sent by swagger, while taking care that both http & https are handled appropriately. I would like for us to go with this approach, as I think it would be more likely to be solved as such for an application used by clients.
     - Disable csrf.

pom.xml
dbernatt: But then I cant use the `@Slf4j` annotation, because its from lombok dependency. In my pov for logging purposes its harmless.

dbernatt: And its rly annoying to write every time that boiler plate static LOGGER attribute.

dbernatt: I don't get it. I assumed that we don't use lombok on models bc its dangerous, but for logging why can't we use it?

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
dbernatt: But it's not clear why not use it?

pom.xml
nymphadora-tonks-team: As mentioned previously on slack, we want to discourage the use of lombok, as we want the focus to be on learning to write good code. This involves both the actual code part (for lack of a better term) and on the configs that are used.
For example here, there are multiple things that lombok will do in the background:
- add all the configs needed in order for the logging framework to work as expected
- choose an implementation
- create the logger entity in the class (I have not used field/constant on purpose in order to not reveal how it would be defined otherwise)

As such, there are a few steps that lombok will provide out of the box, that were skipped otherwise.
Therefore, I would like to encourage setting up the logging part without lombok, in order to see how that would be done.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
nymphadora-tonks-team: Answered here: https://github.com/nymphadora-tonks-team/booking-jb/pull/20#discussion_r749681612

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
```diff
@@ -24,6 +24,8 @@ public void configure(final HttpSecurity http) throws Exception {
                 .and()
                 .logout()
                 .permitAll();
+        // temporal solution
+        http.csrf().disable();
```
nymphadora-tonks-team: As discussed previously, here are the steps that you can follow in order to see how csrf could be enabled
- Google someting like `spring boot swagger post 403`
- The first page for results was this one: https://stackoverflow.com/questions/64837274/swagger-post-return-403-forbidden-spring-boot-spring-security
- Afterwards, you can search how to enable csrf (since above we saw that it is the cause for the issue), by googling something like: `spring boot swagger csrf enable`
- This led to having this `https://stackoverflow.com/questions/61956227/csrf-support-in-springdoc-openapi-swagger-ui` as first result
- Check what the option does & enable it if suitable & test if it works.

nymphadora-tonks-team (commented):
Is the Pull Request ready for review? Or are there some issues that you are still looking into?

TeoFurcovici (commented):
Review done

pom.xml
dbernatt: Fixed. New code available.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
dbernatt: Fixed. New code available.

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
dbernatt: I managed to make it work with csrf enabled in the browser with swagger. Only, I can't make the call from postman.

dbernatt:
I finished with the code. It's ready for review.

loredanaiamnitchi (approved):

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
nymphadora-tonks-team: Can you make the call from postman with the current config/setup?

src/main/java/com/bootcamp/demo/FirebaseInitializer.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.annotation.PostConstruct;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+
+import static com.google.auth.oauth2.GoogleCredentials.fromStream;
+import static com.google.firebase.FirebaseApp.initializeApp;
+import static java.lang.System.getProperty;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+@Service
```
nymphadora-tonks-team: Is this a service, or would it rather be a configuration class, since it only initialises the connection? I think this would be more appropriate to be done in the config class where the firestoreDb bean is also defined.

src/main/java/com/bootcamp/demo/firebase/FirebaseConfig.java
```diff
@@ -0,0 +1,14 @@
+package com.bootcamp.demo.firebase;
```
nymphadora-tonks-team: Not sure if the name of the package should be "firebase" or sth like "repo".

src/main/java/com/bootcamp/demo/firebase/FirebaseController.java
```diff
@@ -30,18 +21,11 @@
 @RestController
 @RequestMapping(path = "/firebase", produces = APPLICATION_JSON_VALUE)
 public class FirebaseController {
```
nymphadora-tonks-team, line 23: Would it be better to have all controller in the same package, to keep the layering?

src/main/java/com/bootcamp/demo/firebase/FirebaseController.java
```diff
@@ -1,25 +1,16 @@
-package com.bootcamp.demo;
+package com.bootcamp.demo.firebase;
```
nymphadora-tonks-team: Would it be better to have all controller in the same package, to keep the layering?

src/main/java/com/bootcamp/demo/firebase/FirebaseController.java
```diff
@@ -30,18 +21,11 @@
 @RestController
```
nymphadora-tonks-team: This would is for the line above, but github does not allow adding comments there.
Check if it makes sense to keep the `@ConditionalOnProperty` annotation with the current setup.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @GetMapping
+    public Set<Scooter> getScooters() {
+        LOGGER.info("GET SCOOTERS - API endpoint invoked");
+        return scooterService.findAllScooters();
+    }
+
+    @GetMapping("/{scooterId}")
+    public Scooter getScooterById(@PathVariable(value = "scooterId") String scooterId) {
+        LOGGER.info("GET SCOOTER BY ID - API endpoint invoked");
```
nymphadora-tonks-team: As a general note for logging, it is usually best to try to include details which would help differentiate the logs from others. For example here, you could log the scooterId, in order to be able to see which call was for it (for example in case it would fail for one id, you could easily search afterwards for it and ease the identification of the cause).

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,42 @@
...
+        return scooterService.findScooterById(scooterId);
+    }
+
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public void postScooter(@RequestBody final Scooter scooter) {
+        LOGGER.info("POST SCOOTER - API endpoint invoked");
+        scooter.getBattery().setStatus();
```
nymphadora-tonks-team: Why does this need to be called explicitly, since it is also done in the constructor?

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,13 @@
+package com.bootcamp.demo.service;
+
+import com.bootcamp.demo.model.Scooter;
+
+import java.util.Set;
+
+public interface ScooterService {
```
nymphadora-tonks-team: Not sure if we need the interfaces, but not opposed to having it as such either. Have you discussed as a team which approach would be better? With or without interfaces?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String COLLECTION_SCOOTERS_PATH = "bookings/databases/scooters";
+    private final Firestore db;
+
+    private ScooterServiceImpl(Firestore db) {
+        this.db = db;
+    }
+
+    @Override
+    public Scooter findScooterById(String scooterId) {
+        LOGGER.info(String.format("FIND SCOOTER BY ID - service function invoked: scooterId = %s", scooterId));
```
nymphadora-tonks-team: As a note, you do not need to do the formatting using `String.format`. You can make use of the built in formatting functionality of the logger, with slightly different placeholders. For example this could be written as:
```
LOGGER.info("FIND SCOOTER BY ID - service function invoked: scooterId = {}", scooterId);
```

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+                    .get();
+
+            if (!scooter.exists()) {
+                LOGGER.error(String.format("Scooter with id = %s", scooterId));
+                throw new ItemNotFoundException();
+            }
+
+            return scooter.toObject(Scooter.class);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error(e.getMessage());
```
nymphadora-tonks-team: As a note, when logging only the message you do not have all the details as you would if the stacktrace would be logged. Also, should the id be logged as well to identify which request it failed?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new ServiceException();
```
nymphadora-tonks-team: If you have decided to throw an exception from the service, you should handle it in controller. Alternatively, you could handle this here without propagating the exceptions. Should I provide more details for both options?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    public Set<Scooter> findAllScooters() {
+        LOGGER.info("FIND ALL SCOOTER - service function invoked");
+
+        try {
+            Iterable<QueryDocumentSnapshot> scooters = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .get()
+                    .get()
+                    .getDocuments();
+
+            LOGGER.info("Found scooters successfully");
```
nymphadora-tonks-team: Not sure if this log is relevant.

src/main/java/com/bootcamp/demo/service/assembler/ScooterAssembler.java
```diff
@@ -0,0 +1,11 @@
+package com.bootcamp.demo.service.assembler;
+
+import com.bootcamp.demo.model.Scooter;
+import com.google.cloud.firestore.QueryDocumentSnapshot;
+
+public class ScooterAssembler {
```
nymphadora-tonks-team: I'm not sure if we need this class since the mapping is straightforward and could be called inline.

src/main/java/com/bootcamp/demo/service/exception/ItemNotFoundException.java
```diff
@@ -0,0 +1,11 @@
+package com.bootcamp.demo.service.exception;
+
+public class ItemNotFoundException extends RuntimeException {
+    public ItemNotFoundException() {
```
nymphadora-tonks-team: When creating exceptions, it is usually best to use custom message which will outline the cause of the failure. You can find some details about this here: https://www.baeldung.com/java-new-custom-exception

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    public Scooter findScooterById(String scooterId) {
+        LOGGER.info(String.format("FIND SCOOTER BY ID - service function invoked: scooterId = %s", scooterId));
+
+        try {
+            DocumentSnapshot scooter = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .document(scooterId)
+                    .get()
+                    .get();
+
+            if (!scooter.exists()) {
```
nymphadora-tonks-team: Could you highlight why you have chosen to throw an exception for this?

src/main/java/com/bootcamp/demo/FirebaseInitializer.java
dbernatt: Moved the initialization to FirebaseConfig. I agree.

src/main/java/com/bootcamp/demo/firebase/FirebaseConfig.java
dbernatt: Agree.

src/main/java/com/bootcamp/demo/firebase/FirebaseController.java
dbernatt: The restapi package intial purpose was to store the controllers for each model, but we can add the firebase controller as well; its not a problem for me. :)

src/main/java/com/bootcamp/demo/firebase/FirebaseController.java
dbernatt: I read more about this `@ConditionalOnProperty` annotation on the official page and concluded that its not necessary to keep it, because we don't define the firebaseKey in the `appliucation.properties`

dbernatt: Fixed.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
dbernatt: Good point. Fixed.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
dbernatt: Forgot to remove it after the refactor. True, no need to call explicitly again.

src/main/java/com/bootcamp/demo/service/ScooterService.java
dbernatt: This solution enables to switch between implementations. For example, if in the future we want to change the database to a different one, then we just create an another Implementation for that database. Makes the transition more flexible, like in the factory design pattern.

dbernatt: This was my idea. The team seemed to accept it. :). This is the first comment/question on this subject. :)

dbernatt: They used this code in their branches as well.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
dbernatt: True. FIxed.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
dbernatt: Removed.

src/main/java/com/bootcamp/demo/service/assembler/ScooterAssembler.java
dbernatt: For the moment we don't have complex object transformations so I agree its not required.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
dbernatt: I fixed with global exception handlers.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
dbernatt: Fixed. Added the item identifier to the LOGGER + the stacktrace is logged in the LOGGER as well.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
dbernatt: I decided to respond with an exception if the scooter is missing instead of returning null. For me, this is more descriptive response and easier to handle in the controller.

src/main/java/com/bootcamp/demo/service/exception/ItemNotFoundException.java
dbernatt: This constructor is for when we don't include the custom error message. If we don't include the message a general message will be returned, which is added in the `GlobalExceptionHandler` class `handleItemNotFoundException()` method. The general message is this: "Failed to find item"

loredanaiamnitchi (approved):

src/main/java/com/bootcamp/demo/model/component/ScooterStatus.java
```diff
@@ -10,6 +10,16 @@
         this.value = value;
     }
 
+    public static boolean contains(String test) {
```
tonks-lupin: Is this method used? If you need such a method it can be rewritten in a simpler way:
`        return asList(ScooterStatus.values()).contains(test);`
where you can extract the list creation as a class constant
or:
       ` return EnumSet.of(AVAILABLE, RESERVED).contains(test);`

src/main/java/com/bootcamp/demo/restapi/controller/FirebaseController.java
```diff
@@ -1,54 +1,40 @@
...
+    private FirebaseController(Firestore firestoreDB) {
+        this.firestoreDB = firestoreDB;
     }
 
     /**
      * Returns the path for all collections stored in Firestore
      */
     @GetMapping("/getAllPaths")
     public Set<String> getAllPaths() {
+        LOGGER.info("GET FIREBASE COLLECTIONS - API endpoint invoked");
```
tonks-lupin: These types of logs are redundant.

src/main/java/com/bootcamp/demo/restapi/dto/battery/BatteryRequestDTO.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
```
tonks-lupin: If the spacing does not increase readability, it is redundant.

src/main/java/com/bootcamp/demo/restapi/dto/battery/BatteryRequestDTO.java
```diff
@@ -0,0 +1,52 @@
...
+        final var battery = (BatteryRequestDTO) o;
+
+        return Objects.equal(this.getLevel(), battery.getLevel());
```
tonks-lupin: What is the advantage of using com.google.common.base.Objects.equal vs java.util.Objects.equals or is this just a wrong import?

src/main/java/com/bootcamp/demo/restapi/dto/battery/BatteryRequestDTO.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(level);
```
tonks-lupin: Similar here, usage of com.google.common.base.Objects

src/main/java/com/bootcamp/demo/restapi/dto/scooter/ScooterRequestDTO.java
```diff
@@ -0,0 +1,91 @@
...
+    private String serialNumber;
+
+    @Valid
+    private LocationRequestDTO currentLocation;
+
+    @Valid
+    private BatteryRequestDTO battery;
+
+    @NotEmpty(message = "Please provide a status")
+    @Pattern(regexp = "^(AVAILABLE|RESERVED)$", message = "Status valid values: AVAILABLE or RESERVED")
```
tonks-lupin: Should case sensitivity be enforced?

src/main/java/com/bootcamp/demo/restapi/dto/scooter/ScooterRequestDTO.java
```diff
@@ -0,0 +1,91 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.Pattern;
+import java.util.Objects;
+
+public class ScooterRequestDTO {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ScooterRequestDTO.class);
```
tonks-lupin: Is the logger used?

src/main/java/com/bootcamp/demo/restapi/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.stream.Collectors;
+
+@ControllerAdvice
+public class GlobalExceptionHandler {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(GlobalExceptionHandler.class);
+
+    /*
+        Global exception handlers
+     */
```
tonks-lupin: If documentation does not add value, it is redundant. You could instead document the purpose of the class itself.

src/main/java/com/bootcamp/demo/restapi/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,115 @@
...
+                .getFieldErrors()
+                .stream()
+                .map(DefaultMessageSourceResolvable::getDefaultMessage)
+                .collect(Collectors.toList());
+        String errorMessage = String.join(", ", errors);
+        HttpStatus status = HttpStatus.BAD_REQUEST;
+        return handleExceptionInternally(new ApiError(errorMessage, status.value()), new HttpHeaders(), status);
+    }
+
+    @ExceptionHandler(NoHandlerFoundException.class)
```
tonks-lupin: If you look into NoHandlerFoundException, you will see that it is thrown if throwExceptionIfNoHandlerFound property is set to true, otherwise the default app behaviour is to return 404 status code. 
Does wrapping this into a BadRequest status add any value?

src/main/java/com/bootcamp/demo/restapi/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,115 @@
...
+    private ResponseEntity<ApiError> handleNoHandlerFoundException(NoHandlerFoundException ex, WebRequest request) {
+        String errorMessage = "Invalid URI: " + request.getContextPath();
+        HttpStatus status = HttpStatus.BAD_REQUEST;
+        return handleExceptionInternally(new ApiError(errorMessage, status.value()), new HttpHeaders(), status);
+    }
+
+    @ExceptionHandler(HttpRequestMethodNotSupportedException.class)
```
tonks-lupin: Sounds more like the default 405 status code.

src/main/java/com/bootcamp/demo/restapi/exceptionhandler/ApiError.java
```diff
@@ -0,0 +1,61 @@
+package com.bootcamp.demo.restapi.exceptionhandler;
+
+import java.util.Objects;
+
+public class ApiError {
+
+    private int status;
```
tonks-lupin: Considering http status code is already returned, does this add value? 
You could stick with a simpler version for exception handling.

src/main/java/com/bootcamp/demo/restapi/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,115 @@
...
+        String errorMessage = ex.getMessage().equals("") ? "Failed to find item" : ex.getMessage();
+        HttpStatus status = HttpStatus.NOT_FOUND;
+        return handleExceptionInternally(new ApiError(errorMessage, status.value()), new HttpHeaders(), status);
+    }
+
+    @ExceptionHandler(ServiceException.class)
+    private ResponseEntity<ApiError> handleServiceException(ServiceException ex) {
+        String errorMessage = "Oops. Something went wrong";
+        HttpStatus status = HttpStatus.INTERNAL_SERVER_ERROR;
+        return handleExceptionInternally(new ApiError(errorMessage, status.value()), new HttpHeaders(), status);
```
tonks-lupin: If you do not need headers, you could use
`public ResponseEntity(@Nullable T body, HttpStatus status)` constructor.

src/main/java/com/bootcamp/demo/service/ScooterService.java
tonks-lupin: As mentioned above, this approach can be debated.
In this specific scenario what value does having an interface add, when there is only one possible implementation of the Services?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            ApiFuture<WriteResult> collectionsApiFuture = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .document(scooter.getSerialNumber())
+                    .set(scooter);
+
+            LOGGER.info("CREATE SCOOTER - created successfully. Update time: {}", collectionsApiFuture.get().getUpdateTime().toDate());
+
+            return scooter;
+
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
```
tonks-lupin: Could you highlight the way to reach an IllegalArgumentException here?

src/main/java/com/bootcamp/demo/service/exception/ItemNotFoundException.java
tonks-lupin: In this specific scenario, do not expose a constructor allowing one to create an exception without any useful message.

src/main/java/com/bootcamp/demo/service/exception/ServiceException.java
```diff
@@ -0,0 +1,20 @@
+package com.bootcamp.demo.service.exception;
+
+public class ServiceException extends RuntimeException {
+
+    public ServiceException() {
+        super("");
```
tonks-lupin: Similar as above.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+
+            if (!scooter.exists()) {
+                LOGGER.warn("FIND SCOOTER BY ID - Scooter does not exist! scooterId = {}", scooterId);
+                throw new ItemNotFoundException(String.format("Scooter does not exist. scooterId = %s", scooterId));
+            }
+
+            return scooter.toObject(Scooter.class);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error("FIND SCOOTER BY ID - service function. scooterId = {}", scooterId);
+            LOGGER.error(ExceptionUtils.getStackTrace(e));
```
tonks-lupin: You could log useful exception details in one log message. This applies everywhere.

src/main/java/com/bootcamp/demo/restapi/controller/ScooterController.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @GetMapping("/{scooterId}")
+    public Scooter getScooterById(@PathVariable(value = "scooterId") String scooterId) {
+        LOGGER.info("GET SCOOTER BY ID - API endpoint invoked. scooterId = {}", scooterId);
+        return scooterService.findScooterById(scooterId);
+    }
+
+    @PostMapping
+    @ResponseStatus(HttpStatus.CREATED)
+    public Scooter postScooter(@Valid @RequestBody ScooterRequestDTO scooterRequestDTO) {
```
nymphadora-tonks-team: Notes regarding the use of DTOs, **which should be discussed with the team**
- when using DTOs, they should be used both for input & output (since the idea is to provide another model, potentially simplified, to the client/user)
- discuss with colleagues about an approach for validation, since if the only reason for having DTOs is for validation (and are as result a wrapper over the model class), perhaps sth else might be more suitable

src/main/java/com/bootcamp/demo/restapi/exceptionhandler/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,115 @@
...
+    private ResponseEntity<ApiError> handleMethodArgumentTypeMismatchException(MethodArgumentTypeMismatchException ex) {
+        String errorMessage = ex.getValue() + " has invalid type.";
+        HttpStatus status = HttpStatus.BAD_REQUEST;
+        return handleExceptionInternally(new ApiError(errorMessage, status.value()), new HttpHeaders(), status);
+    }
+
+    @ExceptionHandler(Exception.class)
+    private ResponseEntity<ApiError> handleException(Exception ex) {
+        // Converting the stacktrace to string and writing to the log file
+        LOGGER.error(ExceptionUtils.getStackTrace(ex));
```
nymphadora-tonks-team: Why is conversion to string needed as such?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
nymphadora-tonks-team: Highlight why logging is called twice (and not once). Also, please look at the documentation of the logger's methods since it will provide more details about how it can be used.

src/main/java/com/bootcamp/demo/service/ScooterService.java
dbernatt: [This](https://dimitr.im/spring-interface) article convinced me to why not use it. :) Thanks pointing it out.

dbernatt:
Moved the validation + exception handling + dtos part to a different branch.

dbernatt:
The problem is I can't comment on these, because I moved this code base(validation + exception handling + dtos + delete by id) into the #28 PR. @tonks-lupin @nymphadora-tonks-team

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.Set;
+
+@RestController
+@RequestMapping(path = "/api/scooters")
+public class ScooterController {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ScooterController.class);
```
nymphadora-tonks-team: It looks like this is no longer used. As such, it should be removed.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,58 @@
...
+    private final ScooterService scooterService;
+
+    public ScooterController(ScooterService scooterService) {
+        this.scooterService = scooterService;
+    }
+
+    @GetMapping
+    @ResponseBody
+    public ResponseEntity<Set<Scooter>> getScooters() {
+        Set<Scooter> scooters = null;
```
nymphadora-tonks-team: No need to use an intermediate variable for this approach. You could simplify this to sth like:
```
try {
    return new ResponseEntity<>(scooterService.findAllScooters(), HttpStatus.OK);
} catch (ServiceException e) {
}
```
which is similar to what you have done in `getScooterById`, with the difference that the find call is inlined.

_Just as a note:_
Personally I find it easier to have only one return statement per method, but for that you would need to have sth like
```
Set<Scooter> scooters = null;
var status = HttpStatus.OK;
try {
    scooters = scooterService.findAllScooters();
} catch (ServiceException e) {
    status = HttpStatus.BAD_REQUEST;
}
return new ResponseEntity<>(scooters, status);
```
and in this case it is probably not very intuitive.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,58 @@
...
+            return new ResponseEntity<>(scooter, HttpStatus.OK);
+        } catch (ServiceException e) {
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @PostMapping
+    public ResponseEntity<Object> postScooter(@RequestBody final Scooter scooter) {
+        try {
+            scooter.getBattery().setStatus();
```
nymphadora-tonks-team: Could you outline why it is necessary to call this directly here? Note the following:
- If empty constructor is used for deserialization, then the fields from `Battery` will be set through setters. In this case, the `setLevel` also calls the `setStatus` method.
- If the constructor with all params is used for deserialization, then the `setStatus` method is called in the constructor.

Thus, in both cases, the Battery should be created in a consistent manner & have status set appropriately.
If that is not the case, could you provide details regarding what happens?

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
```diff
@@ -31,6 +31,8 @@ public void configure(final HttpSecurity http) throws Exception {
                 .and()
                 .logout()
                 .permitAll();
+
+        http.csrf().csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse());
```
nymphadora-tonks-team: This is called above as well, so no need to duplicate it here as well. Probably was added when resetting the branch, but just as a note that it can be removed.

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,86 @@
...
+    public Scooter findScooterById(String scooterId) {
+        LOGGER.info("FIND SCOOTER BY ID - service function invoked: scooterId = {}", scooterId);
+
+        try {
+            DocumentSnapshot scooter = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .document(scooterId)
+                    .get()
+                    .get();
+
+            if (!scooter.exists()) {
```
nymphadora-tonks-team: Could you highlight why you have chosen to throw an exception for this (rather than returning empty/null) ?

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(ScooterService.class);
+    private static final String COLLECTION_SCOOTERS_PATH = "bookings/databases/scooters";
+    private final Firestore db;
+
+    private ScooterService(Firestore db) {
+        this.db = db;
+    }
+
+    public Scooter findScooterById(String scooterId) {
+        LOGGER.info("FIND SCOOTER BY ID - service function invoked: scooterId = {}", scooterId);
```
nymphadora-tonks-team: Notes about logging: https://sematext.com/blog/java-logging-best-practices/ . You can find more details here regarding best practices for logging, including recommended approach for content.

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,86 @@
...
+                    .get();
+
+            if (!scooter.exists()) {
+                LOGGER.error("FIND SCOOTER BY ID - service function: Scooter does not exists! scooterId = {}", scooterId);
+                throw new IllegalArgumentException();
+            }
+
+            return scooter.toObject(Scooter.class);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.error("FIND SCOOTER BY ID - service function: scooterId = {}.\n Error message: {}.", scooterId, e.getMessage(), e);
```
nymphadora-tonks-team: Highlight why is the separate message logging necessary, since you log the exception (which will include both the message & the stacktrace).

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public Set<Scooter> findAllScooters() {
+        try {
+            Iterable<QueryDocumentSnapshot> scooters = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .get()
+                    .get()
+                    .getDocuments();
+
+            return StreamSupport
+                    .stream(scooters.spliterator(), false)
```
nymphadora-tonks-team: Note that you could have used `.stream` directly on the documents' list, thus resulting in the following:
```
return db.collection(COLLECTION_SCOOTERS_PATH)
        .get()
        .get()
        .getDocuments()
        .stream()
        .map(queryDocumentSnapshot -> queryDocumentSnapshot.toObject(Scooter.class))
        .collect(Collectors.toSet());
```

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public void createScooter(final Scooter scooter) {
+        LOGGER.info("CREATE SCOOTER - service function invoked: scooter = {}", scooter);
+
+        try {
+            ApiFuture<WriteResult> collectionsApiFuture = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .document(scooter.getSerialNumber())
+                    .set(scooter);
+
+            String update_time = collectionsApiFuture.get().getUpdateTime().toDate().toString();
```
nymphadora-tonks-team: Note that for java variables, snake case naming is used.

src/main/resources/application.properties
```diff
@@ -3,4 +3,14 @@ spring.application.name=booking-java-bootcamp
 springdoc.swagger-ui.enabled=true
 springdoc.swagger-ui.path=/swagger.html
 // enabling CSRF support, while using standard header names
-springdoc.swagger-ui.csrf.enabled=true
\ No newline at end of file
+springdoc.swagger-ui.csrf.enabled=true
+
+// enabling access logs
```
nymphadora-tonks-team: Note that comments in properties files are added with #, not //.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
dbernatt: Here is what happens:
Case 1: I explicitly call the `scooter.getBattery().setStatus();` in `postScooter(Scooter scooter)`
Scooter in:
```
{
  "serialNumber": "string",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 0,
    "status": "HIGH"
  },
  "status": "AVAILABLE"
}
```
Result: 
```
{
  "serialNumber": "string",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 0,
    "status": "LOW"
  },
  "status": "AVAILABLE"
}
```
Case 2: Without the  `scooter.getBattery().setStatus();` call
Scooter in:
```
{
  "serialNumber": "scooter-id-1",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 0,
    "status": "HIGH"
  },
  "status": "AVAILABLE"
}
```
Result:
```
{
  "serialNumber": "scooter-id-1",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 0,
    "status": "HIGH"
  },
  "status": "AVAILABLE"
}
```
So the setStatus() or setLevel() were never called :)).

Case3: Without the  `scooter.getBattery().setStatus();` call, but without battery status input.
```
{
  "serialNumber": "string",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 100
  },
  "status": "AVAILABLE"
}
```
Result:
```
{
  "serialNumber": "string",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 100,
    "status": "HIGH"
  },
  "status": "AVAILABLE"
}
```

src/main/java/com/bootcamp/demo/service/ScooterService.java
dbernatt: This article helped me a lot because I couldn't find this best practices examples before. Only have 1 question:
What about masking sensitive information (like the scooterId). Should I create a converter which masks the information or I can leave it for now?

src/main/java/com/bootcamp/demo/service/ScooterService.java
dbernatt: Yeah, now I know, after reading about how to log messages properly, [link](https://sematext.com/blog/java-logging-best-practices/). Its fixed now.

src/main/java/com/bootcamp/demo/service/ScooterService.java
dbernatt: This solution was used in case we have custom exceptions with proper exception handlers which will automatically catch and respond back to the caller. But in our case we don't have yet these kind of handlers so I can just return a null. np.

loredanaiamnitchi (approved):

src/main/java/com/bootcamp/demo/service/ScooterService.java
nymphadora-tonks-team: Usually, when possible, it is good to log the information that uniquely identifies the entry. For example, if you would have an exception on scooter's retrieval, it would be helpful to have the id in the log, since then you could look in the db and check if anything was changed to its structure in a way that prevents deserialization. Also, this is not sensitive information.
If it would be for a user on the other hand, it would be preferable to log the id and not the email, since the later one would be covered by GDPR. As a note, if you would need to log such information (sensitive), you should convert it in such a way that it cannot be converted back to the original value (consider here why for example hashing could be used and encryption not; you can find an overview of these approaches here: https://www.geeksforgeeks.org/difference-between-hashing-and-encryption/)
In conclusion, it is ok to use the id.

dbernatt: Ok. Thanks for the detailed answer/suggestions.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
nymphadora-tonks-team: And have you looked into why this happens (through debugging)?
From what you have outlined, it looks like it is caused by the fact that `setLevel` (which would set the status to LOW) is called before `setStatus` (which will set the status to the incorrectly provided value). Although having afterwards the setter called here explicitly solves the problem for now, it can lead to subtle future issues & introduces an unwanted coupling between creation & always calling this method afterwards.
What do you think about the following as a solution to not need this: Have the `Battery` class with the 2 fields that it has now, but not expose serialise status for it? This would result in having the following serialised structure (as json):
```
{
  "serialNumber": "string",
  "currentLocation": {
    "latitude": 0,
    "longitude": 0
  },
  "battery": {
    "level": 0
  },
  "status": "AVAILABLE"
}
```
but when operating with the model in the app, we would still have the status (and can be used when needed).
This can be easily achieved by annotating the `getStatus` with `@JsonIgnore`.

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,77 @@
...
+    public Scooter findScooterById(String scooterId) {
+        try {
+            DocumentSnapshot scooter = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .document(scooterId)
+                    .get()
+                    .get();
+
+            if (!scooter.exists()) {
+                LOGGER.info("FIND SCOOTER BY ID - service function: Scooter does not exists! scooterId = {}", scooterId);
+                return null;
```
nymphadora-tonks-team: In case this log is not needed, you could simplify as 
```
return db.collection(COLLECTION_SCOOTERS_PATH)
        .document(scooterId)
        .get()
        .get()
        .toObject(Scooter.class)
```
Also as a note that calling that directly would not throw an exception when the document does not exist.

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public void createScooter(final Scooter scooter) {
+        try {
+            ApiFuture<WriteResult> collectionsApiFuture = db.collection(COLLECTION_SCOOTERS_PATH)
+                    .document(scooter.getSerialNumber())
+                    .set(scooter);
+
+            String updateTime = collectionsApiFuture.get().getUpdateTime().toDate().toString();
+
+            LOGGER.info("CREATE SCOOTER - created successfully. scooter = {}. Update_time: {}", scooter, updateTime);
```
nymphadora-tonks-team: As you have seen in the article provided in a previous comment, there is a balance between logging enough info in order to be relevant and understand what happened & logging too many details. For example for this log & the below one, how about the following (applies for the other logs as well, but taken this as example since it had logs for success & failure as well):
- For success: `LOGGER.info("Created scooter with id = {} at = {}", scooter, updateTime);`
- For failure: `LOGGER.error("Failed to create scooter with id = {}.", scooter, e);`

Note that those provide the same details, but omit the prefix that you have added, since it is not necessarily relevant, as usually there is a single point where crud operations exist in an app.

dbernatt: Agree.

src/main/java/com/bootcamp/demo/service/ScooterService.java
dbernatt: FIxed.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
dbernatt: That solved partially the problem, because when I make a GET request then the battery status is not included in the response, which is not what I wanted tho. (its ignored in both direction).

nymphadora-tonks-team: Indeed, it is not shown, but wondering if it is shown/needed currently. If it is, we could have this limitation/coupling for the moment, but we should probably document it, so that for possible further updates for a scooter, this is taken into account.

nymphadora-tonks-team (approved):
I think the only part left is to decide if a battery's status should be included in the serialised version or not.

PR #21, opened 2021-11-12
View available scooters 

### LINKED ISSUE #4: Book a scooter - view available scooters by location and charge (M)
+- approximate 
filter available ones
batteryLevelEnum/thresholding 0-30% red 30-50% yellow 50%+ green

loredanaiamnitchi:
Implementation for https://github.com/nymphadora-tonks-team/booking-jb/issues/4. 
Starting from Bernat's code from the [feature / bernat-add-scooter- # 18 branch](https://github.com/nymphadora-tonks-team/booking-jb/tree/feature/bernat-add-scooter-%2318), I added methods corresponding to my task in the ScooterService, ScooterServiceImpl and ScooterController classes.


src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,65 @@
...
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public void postScooter(@RequestBody final Scooter scooter) {
+        LOGGER.info("POST SCOOTER - API endpoint invoked");
+        scooterService.createScooter(scooter);
+    }
+
+    @GetMapping("/available")
+    public Set<Scooter> getAvailableScooters(){
+        Location userLocation = new Location();
```
dbernatt: The user location comes from the user, when he calls this function. Its not a default value. You need to include the userlocation in the call with `pathvariables`.  And the desired radius where we need to search the scooters.

src/main/java/com/bootcamp/demo/model/util/Battery.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || this.getClass() != o.getClass()) {
+            return false;
+        }
+        final var battery = (Battery) o;
+        return this.getLevel().equals(battery.getLevel())
```
dbernatt: Same here. -> Objects.equals()

src/main/java/com/bootcamp/demo/model/Scooter.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || this.getClass() != o.getClass()) {
+            return false;
+        }
+        final var scooter = (Scooter) o;
+        return this.getSerialNumber().equals(scooter.getSerialNumber())
```
dbernatt: You need to change the `obj1.equals(obj2)` to `Objects.equals(obj1, obj2)`, because of null potential null values.

src/main/java/com/bootcamp/demo/model/util/Location.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || this.getClass() != o.getClass()) {
+            return false;
+        }
+        final var location = (Location) o;
+        return this.getLatitude().equals(location.getLatitude())
```
dbernatt: -> Objects,equals()

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
...
+        this.firstName = firstName;
+    }
+
+    public void setLastName(String lastName) {
+        this.lastName = lastName;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
```
dbernatt: Its a good practice to use `{` `}` even for 1 statement in the if block.

src/main/java/com/bootcamp/demo/models/User.java
```diff
@@ -0,0 +1,72 @@
...
+        if (obj == null || getClass() != obj.getClass()) return false;
+        final User other = (User) obj;
+        return Objects.equals(userId, other.userId);
```
dbernatt: Missing other attribute checks. The equals method checks all the attributes. Not just the id.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            String update_time = collectionsApiFuture.get().getUpdateTime().toDate().toString();
+
+            LOGGER.info("SCOOTER create successfully. Update_time: " + update_time);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error(e.getMessage());
+            throw new ServiceException();
+        }
+    }
+
+    private boolean approximateLocation(Location userLocation, Location scooterLocation) {
```
dbernatt: The 0.01 radius is approx. 1.573 km radius. 1.573 km in a city is not precise in my pov. My suggestion is to add a new attribute which will asks the required approximate location radius in meters from the user. For ex: 100m. 250m, 500m, 1000 m. 2000m radius etc.
So the asks exactly what is the desired radius where he/she needs to search available scooters. 
50m radius: ->lat/lon: 0.000318
100m radius: -> lat/lon:  0.000636
250m radius: -> lat/lon: 0.00159
500m radius: -> lat/lon:  0.00318
1000m radius: -> lat/lon: 0.00636
and so on....
We can discuss this in slack channel if you want with the team.

dbernatt (changes requested):
Added my comments and requested some changes. Checkout it out. :)

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+                && (userLocation.getLatitude() - 0.01 < scooterLocation.getLatitude())) {
+                return true;
+            }
+        }
+        System.out.println("The scooter is not nearby");
+        return false;
+    }
+
+    @Override
+    public Set<Scooter> getAvailableScooters(Location userLocation) {
```
dbernatt: Same here we need to return the available scooters in an ordered manner. Here is the explanation. https://stackoverflow.com/questions/326634/sorting-on-the-server-or-on-the-client/326641

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        LOGGER.info("GET AVAILABLE SCOOTERS - service function invoked");
+        Set<Scooter> allScooters = findAllScooters();
```
dbernatt: Usually the service methods are independent from each other. We may need to create a repository where we implement all the methods we need to interact with the database.

dbernatt (changes requested):

TeoFurcovici (approved):

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
dbernatt: But you can leave it now.

lombok.config
```diff
@@ -0,0 +1,4 @@
+# stop lombok from looking at parent directories for more configuration files
```
nymphadora-tonks-team: As discussed previously on slack (and on some other Pull Requests), you should not use lombok as we would like the focus to be on the code that you write and not on code generation.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
nymphadora-tonks-team: Small note, the location can also be included in the body, since you could also use POST instead of GET.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public void postScooter(@RequestBody final Scooter scooter) {
+        LOGGER.info("POST SCOOTER - API endpoint invoked");
+        scooterService.createScooter(scooter);
+    }
+
+    @GetMapping("/available")
+    public Set<Scooter> getAvailableScooters(){
```
nymphadora-tonks-team: Try to format the code, as it helps to keep the same structure. Here the missing space is minor, but it is good practice to get used to doing this.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        Set<Scooter> availableScooters = new HashSet<>();
+        for (Scooter scooter : allScooters) {
```
nymphadora-tonks-team: You could try rewriting the for using stream functionalities. If you would like, I could reference some articles/urls where you would find details about this. 
Optional, but I thought it might be a nice exercise since perhaps you have not used that very much.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            if(Objects.equals(scooter.getStatus().getValue(), "AVAILABLE")) {
```
nymphadora-tonks-team: You should use `==` when comparing enums. You can find more details regarding this in https://www.baeldung.com/a-guide-to-java-enums in the section `3. Comparing Enum Types Using “==” Operator`.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+
+            LOGGER.info("SCOOTER create successfully. Update_time: " + update_time);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error(e.getMessage());
+            throw new ServiceException();
+        }
+    }
+
+    private boolean approximateLocation(Location userLocation, Location scooterLocation) {
+        if (scooterLocation.getLongitude() <= userLocation.getLongitude() + 0.01
```
nymphadora-tonks-team: Since the conditions from if are very similar, wondering if we could extract this to another method which would fo the comparison for a coordinate & apply it for both. This would result in sth like isNearby(user long/lat, scooter long/lat) (with a better name probably), which would be called for both. 
If that will be done, you could also use a ternary for this.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
nymphadora-tonks-team: Usually it is best to use methods names which are intuitive and help understand what the method does. For example here, does `approximateLocation` work best, or would sth like is/areNearby, is/areInProximity fit better?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            && (scooterLocation.getLongitude() > userLocation.getLongitude() - 0.01 )) {
+            if ((userLocation.getLatitude() + 0.01 >= scooterLocation.getLatitude())
+                && (userLocation.getLatitude() - 0.01 < scooterLocation.getLatitude())) {
+                return true;
+            }
+        }
+        System.out.println("The scooter is not nearby");
```
nymphadora-tonks-team: Usually it is best to use logging instead of this. The main reason is that logging can be configured more specifically and to suit deployment (e.g. you would log in a file for a deployed application, rather than in a console).

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -40,10 +40,10 @@ public void postScooter(@RequestBody final Scooter scooter) {
         scooterService.createScooter(scooter);
     }
 
-    @PostMapping("/available")
+    @PostMapping("/available/{searchRadius}")
```
dbernatt, line 52: Why use POST instead of GET? You don't change anything on the server and its not sensitive information (searchRadius + location).

dbernatt (changes requested):

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
loredanaiamnitchi: Because I need RequestBody for location and HTTP's GET method does not include a request body as part of the spec.

dbernatt (approved):

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -0,0 +1,49 @@
...
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public void postScooter(@RequestBody final Scooter scooter) {
+        LOGGER.info("POST SCOOTER - API endpoint invoked");
+        scooterService.createScooter(scooter);
+    }
+
+    @PostMapping("/available/{searchRadius}")
+    @ResponseStatus(value = HttpStatus.OK)
+    public Set<Scooter> getAvailableScooters(@PathVariable(value = "searchRadius") Double searchRadius, @RequestBody final Location selectedLocation) {
+        LOGGER.info("POST SCOOTER - API endpoint invoked");
```
tonks-lupin: These types of logs are redundant.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            String update_time = collectionsApiFuture.get().getUpdateTime().toDate().toString();
+
+            LOGGER.info("SCOOTER create successfully. Update_time: " + update_time);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error(e.getMessage());
+            throw new ServiceException();
+        }
+    }
+
+    private boolean compareCoordinates(Double selectedCoord, Double scooterCoord, Double convertedRadius){
```
tonks-lupin: You could find a better name for this method, maybe something like `isWithinRadius`?

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        return (scooterCoord <= selectedCoord + convertedRadius)
+                && (scooterCoord > selectedCoord - convertedRadius);
+    }
+    
+    private boolean isNearby(Location selectedLocation, Location scooterLocation, Double searchRadius) {
+        Double convertedRadius = searchRadius * 0.00000636;  // 1m=0.00000636 lat/lon
+        return (compareCoordinates(selectedLocation.getLongitude(), scooterLocation.getLongitude(), convertedRadius)
```
tonks-lupin: Embracing parentheses are redundant.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    private boolean compareCoordinates(Double selectedCoord, Double scooterCoord, Double convertedRadius){
+        return (scooterCoord <= selectedCoord + convertedRadius)
+                && (scooterCoord > selectedCoord - convertedRadius);
+    }
+    
+    private boolean isNearby(Location selectedLocation, Location scooterLocation, Double searchRadius) {
+        Double convertedRadius = searchRadius * 0.00000636;  // 1m=0.00000636 lat/lon
```
tonks-lupin: Rather than a short inline comment, you could add Javadoc to the method itself, assume a colleague has to understand it on 1st interaction with the code.

tonks-lupin (approved):
Great job! Approved with the minor improvements handling. Take care with merging since your code depends on another branch.

src/main/java/com/bootcamp/demo/service/ScooterServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        return (compareCoordinates(selectedLocation.getLongitude(), scooterLocation.getLongitude(), convertedRadius)
+                && compareCoordinates(selectedLocation.getLatitude(), scooterLocation.getLatitude(), convertedRadius));
+    }
+
+    @Override
+    public Set<Scooter> getAvailableScooters(Location selectedLocation, Double searchRadius) {
+        LOGGER.info("GET AVAILABLE SCOOTERS - service function invoked");
+        Set<Scooter> allScooters = findAllScooters();
+
+        Set<Scooter> availableScooters = allScooters.stream()
```
nymphadora-tonks-team: You could make the call to  `findAllScooters` here, since the variable is not needed for anything else.

nymphadora-tonks-team (approved):

birleadrian (approved):

PR #22, opened 2021-11-13
View booking's history

### LINKED ISSUE #8: View my bookings' history (M)

mariatirlea:
Implementation for issue View my bookings' history #8. 
Created service classes with methods for : 
                             - retrieving bookings by userID
                             - retrieving booking by bookingID
                             - retrieving all bookings from the db collection
                             - adding one booking 
I tested the operations with methods implemented in FirebaseController.java but I got a 403 response when trying them on Swagger.


src/main/java/com/bootcamp/demo/FirebaseController.java
```diff
@@ -42,16 +53,67 @@ private void initFirestore() throws IOException {
...
     @GetMapping("/getAllPaths")
     public Set<String> getAllPaths() {
-        return StreamSupport.stream(firestoreDB.listCollections().spliterator(), false)
+        return stream(firestoreDB.listCollections().spliterator(), false)
                 .map(CollectionReference::getPath)
                 .collect(Collectors.toUnmodifiableSet());
 
     }
+
+    @PostMapping("/createBooking")
```
dbernatt, line 70: The class is the firebase controller! Usually we make a controller for each model, which is responsible every call you make for that model. So its better to create a separate controller for the booking collection which is only responsible for the booking rest api calls.

dbernatt: You can checkout how to do that in the create/retrieve a scooter PR. [link](https://github.com/nymphadora-tonks-team/booking-jb/pull/20/files)

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.logging.Logger;
+
+@Service
+public class BookingService implements IBookingService {
+    private static final Logger LOGGER = Logger.getLogger(BookingService.class.getName());
+    private static final String COLLECTION_PATH = "bookings";
+
+    private Firestore db;
+
+    public void setDb(Firestore db) {
```
dbernatt, line 24: We inject the firestore dependency via private constructor.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+import java.util.logging.Logger;
+
+@Service
+public class BookingService implements IBookingService {
+    private static final Logger LOGGER = Logger.getLogger(BookingService.class.getName());
+    private static final String COLLECTION_PATH = "bookings";
+
+    private Firestore db;
```
dbernatt, line 22: This should be final.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+    public Booking getBookingByID(final UUID bookingId) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (bookingId == null) throw new IllegalArgumentException();
+        DocumentReference documentReference =  db.collection(COLLECTION_PATH)
+                .document(bookingId.toString());
+        ApiFuture<DocumentSnapshot> future = documentReference.get();
+        DocumentSnapshot document = future.get();
+        Booking booking;
+        if(document.exists()){
+            booking = document.toObject(Booking.class);
+            return booking;
```
dbernatt, line 41: you don't need to create a new booking object. You can simply return the `document.toObject(Booking.class)`. Like you did in the `getBookings()` method

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Retrieves a Booking entity by its id.
+     */
+    @Override
+    public Booking getBookingByID(final UUID bookingId) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (bookingId == null) throw new IllegalArgumentException();
+        DocumentReference documentReference =  db.collection(COLLECTION_PATH)
+                .document(bookingId.toString());
+        ApiFuture<DocumentSnapshot> future = documentReference.get();
+        DocumentSnapshot document = future.get();
```
dbernatt, line 37: You can chain together the calls into the `DocumentSnapshot ` object. So you don't create the `documentReference/future` variables.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    @Override
+    public Set<Booking> getBookings(final UUID userId) throws ExecutionException, InterruptedException, IllegalArgumentException  {
+        if (userId == null) throw new IllegalArgumentException();
+        Set<Booking> userBookings;
+        userBookings = db.collection(COLLECTION_PATH)
+                .whereEqualTo("accountId", userId).get().get()
+                .getDocuments()
+                .stream().map(d -> d.toObject(Booking.class))
+                .collect(Collectors.toSet());
```
dbernatt, line 57: But you need to add an orderBy() method to this chain to order the result by an attribute, like startdate or enddate.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+        return userBookings;
+    }
+
+    /**
+     * Retrieves all entities of type Booking.
+     */
+    @Override
+    public Set<Booking> getAllBookings() throws ExecutionException, InterruptedException {
```
dbernatt, line 65: The reason why we don't want to Set<Booking>, because the set returns the booking history with unordered manner. Usually the user wants to list their history with chronological order. For this problem we usually use LinkedHashSet<>.
https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashSet.html

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+        Set<Booking> allBookings;
+        allBookings = db.collection(COLLECTION_PATH)
+                .get()
+                .get()
+                .getDocuments()
+                .stream().map(d -> d.toObject(Booking.class))
+                .collect(Collectors.toSet());
```
dbernatt, line 72: But you need to add an orderBy() method to this chain to order the result by an attribute, like startdate or enddate.

dbernatt (changes requested):

TeoFurcovici (approved):

src/main/java/com/bootcamp/demo/FirebaseController.java
```diff
@@ -42,16 +53,67 @@ private void initFirestore() throws IOException {
...
+    public ResponseEntity<Object> createBooking(final Booking booking){
```
nymphadora-tonks-team, line 71: Try to format the code, as it helps to keep the same structure. Here the missing space is minor, but it is good practice to get used to doing this.

src/main/java/com/bootcamp/demo/FirebaseController.java
nymphadora-tonks-team: I also think it would be better to move this to a dedicated controller. This would also follow best practice principles.

src/main/java/com/bootcamp/demo/FirebaseController.java
```diff
@@ -42,16 +53,67 @@ private void initFirestore() throws IOException {
...
+        try {
+            LOGGER.info("Booking created successfully. Update time: " + bookingService.createBooking(booking));
+            return SUCCESS_RESPONSE;
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            e.printStackTrace();
```
nymphadora-tonks-team, line 76: You could use logger for logging exceptions as well. If you did that, all details will be logged using the same format and in the same place.

src/main/java/com/bootcamp/demo/FirebaseController.java
nymphadora-tonks-team: Also, should this be marked with `@RequestBody`?

src/main/java/com/bootcamp/demo/FirebaseController.java
```diff
@@ -42,16 +53,67 @@ private void initFirestore() throws IOException {
...
+            return FAILURE_RESPONSE;
+        }
+    }
+
+    @GetMapping("/getBookingsByUserId")
+    @ResponseBody
+    public ResponseEntity<Set<Booking>> getBookingsByUserId(final UUID userId) {
```
nymphadora-tonks-team, line 83: For `@GetMapping`, you should query string params/path variables. If you would like to provide this in the body, you should use `@PostMapping`. You can find an overview for this here: https://www.baeldung.com/spring-new-requestmapping-shortcuts

src/main/java/com/bootcamp/demo/FirebaseController.java
```diff
@@ -30,7 +37,11 @@
 @RestController
 @RequestMapping(path = "/firebase", produces = APPLICATION_JSON_VALUE)
 public class FirebaseController {
+    private static final ResponseEntity<Object> SUCCESS_RESPONSE = new ResponseEntity<>(HttpStatus.OK);
+    private static final ResponseEntity<Object> FAILURE_RESPONSE = new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+    private static final Logger LOGGER = Logger.getLogger(FirebaseController.class.getName());
     private Firestore firestoreDB;
+    private final BookingService bookingService = new BookingService();
```
nymphadora-tonks-team, line 44: Since `BookingService` is a bean, you should not instantiate it directly. Rather, you should inject it. Let me know if more details are needed for this.

src/main/java/com/bootcamp/demo/FirebaseController.java
```diff
@@ -42,16 +53,67 @@ private void initFirestore() throws IOException {
         initializeApp(options);
 
         firestoreDB = getFirestore();
+        bookingService.setDb(firestoreDB);
```
nymphadora-tonks-team, line 56: This will not be needed when using injection.

src/main/java/com/bootcamp/demo/service/BookingService.java
nymphadora-tonks-team: As a note, setter injection can also be used. Should not be invoked directly though.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    @Override
+    public Booking getBookingByID(final UUID bookingId) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (bookingId == null) throw new IllegalArgumentException();
+        DocumentReference documentReference =  db.collection(COLLECTION_PATH)
+                .document(bookingId.toString());
+        ApiFuture<DocumentSnapshot> future = documentReference.get();
+        DocumentSnapshot document = future.get();
+        Booking booking;
+        if(document.exists()){
```
nymphadora-tonks-team, line 39: The explicit check for `exists` is not needed, since, as highlighted in the javadoc of the `toObject` method, it will return null when it does not exist, same as you do explicitly. Javadoc for reference:
```
  /**
   * Returns the contents of the document converted to a POJO or null if the document doesn't exist.
   *
   * @param valueType The Java class to create
   * @return The contents of the document in an object of type T or null if the document doesn't
   *     exist.
   */
  @Nullable
  public <T> T toObject(@Nonnull Class<T> valueType) {

```

src/main/java/com/bootcamp/demo/service/BookingService.java
nymphadora-tonks-team: Highlight also why you think that is necessary.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Retrieves a Booking entity by its user's id.
+     */
+    @Override
+    public Set<Booking> getBookings(final UUID userId) throws ExecutionException, InterruptedException, IllegalArgumentException  {
+        if (userId == null) throw new IllegalArgumentException();
```
nymphadora-tonks-team, line 51: Even when the body of `if` contains only one instruction, it is usually best to add surround it by { }. That is best practice as otherwise it is very easy to add another line here and result in unexpected behaviour.
Also, try to have one instruction/call per line, as it usually improves readability.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+        Set<Booking> userBookings;
+        userBookings = db.collection(COLLECTION_PATH)
```
nymphadora-tonks-team, line 53: You could simply return the result from this call without an intermediate variable, e.g. `return db.collection...`.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+        return allBookings;
+    }
+
+    /**
+     * Creates a new document containing the given Booking type param.
+     */
+    @Override
+    public String createBooking(final Booking booking) throws ExecutionException, InterruptedException,  IllegalArgumentException  {
+        if (booking == null) throw new IllegalArgumentException();
+        ApiFuture<WriteResult> collectionApiFuture  = db.collection(COLLECTION_PATH)
```
nymphadora-tonks-team, line 82: Same here, you could chain all calls and return that result directly, without using an intermediate variable.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,87 @@
...
+        return userBookings;
+    }
+
+    /**
+     * Retrieves all entities of type Booking.
+     */
+    @Override
+    public Set<Booking> getAllBookings() throws ExecutionException, InterruptedException {
+        Set<Booking> allBookings;
+        allBookings = db.collection(COLLECTION_PATH)
```
nymphadora-tonks-team, line 67: Same here, you could simply return the result from this call without an intermediate variable, e.g. `return db.collection....`

PR #25, opened 2021-11-17
Configuration of the Cross site request forgery problem

### LINKED ISSUE #24: Cross site request forgery (csrf) prevention solution
Enabling csrf in the SecurityConfig to solve the 403 forbidden response.

dbernatt:
Added the necessary configuration to solve the csrf problem in the SecurityConfig class.

### DIFF
--- a/src/main/java/com/bootcamp/demo/security/SecurityConfig.java
+++ b/src/main/java/com/bootcamp/demo/security/SecurityConfig.java
@@ -14,13 +14,14 @@ public class SecurityConfig extends WebSecurityConfigurerAdapter {
     @Override
     public void configure(final HttpSecurity http) throws Exception {
         http
+                .csrf().csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse())
+                .and()
                 .authorizeRequests()
                 .antMatchers("/public/**").permitAll()
                 .antMatchers("/actuator/health").permitAll()
                 .antMatchers("/swagger-ui/**").permitAll()
                 .antMatchers("/swagger-resources/**").permitAll()
                 .antMatchers("/webjars/**").permitAll()
-                .antMatchers("/v3/**").permitAll()
                 .anyRequest().authenticated()
                 .and()
                 .formLogin()
@@ -30,8 +31,6 @@ public class SecurityConfig extends WebSecurityConfigurerAdapter {
                 .and()
                 .logout()
                 .permitAll();
-
-        http.csrf().csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse());
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/bootcamp/demo/security/SecurityConfig.java
```diff
@@ -5,16 +5,22 @@
...
 
 @Configuration
 @EnableWebSecurity
 public class SecurityConfig extends WebSecurityConfigurerAdapter {
 
     @Override
     public void configure(final HttpSecurity http) throws Exception {
-        http.authorizeRequests()
+        http
+                .authorizeRequests()
```
nymphadora-tonks-team, line 19: You could have also used
```
http.csrf()
        .csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse())
        .and()
```
to chain the calls.
It is also ok as is, just a note for the alternative.

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
```diff
@@ -5,16 +5,22 @@
...
                 .antMatchers("/public/**").permitAll()
                 .antMatchers("/actuator/health").permitAll()
+                .antMatchers("/swagger-ui/**").permitAll()
+                .antMatchers("/swagger-resources/**").permitAll()
+                .antMatchers("/webjars/**").permitAll()
+                .antMatchers("/v3/**").permitAll()
```
nymphadora-tonks-team: Could you highlight why this path is needed?

src/main/resources/application.properties
```diff
@@ -2,3 +2,5 @@ spring.application.name=booking-java-bootcamp
 
 springdoc.swagger-ui.enabled=true
 springdoc.swagger-ui.path=/swagger.html
+// this is very important configuration, don't remove it! This missing line caused me apprx. 4 hour of pain :))
```
nymphadora-tonks-team: For comments, it is usually best to add a comment in which to describe what this fixes/addresses. As is, when looking over this (regardless of who & when - since you might come back to this after some time and not recall the reasoning for it), it will be hard to understand what it is needed for.

dbernatt: :)). I just added this comment because I wanted to express my pain when I discovered that why the POST call isn't working on swagger. :))

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
dbernatt: Ok

src/main/resources/application.properties
dbernatt: I know the comment is not that helpful but to me that time was so funny :)).

dbernatt: Can we leave it ? Its harmless :)...

src/main/java/com/bootcamp/demo/security/SecurityConfig.java
dbernatt: We don't use this path so yeah, its an unused path. Probably a ctrl+c, ctrl+v leftover from stackoverflow :)).

src/main/resources/application.properties
dbernatt: Fixed.

loredanaiamnitchi (approved):

vladmsn (approved):

TeoFurcovici (approved):

mariatirlea (approved):

nymphadora-tonks-team (approved):

birleadrian (approved):

PR #26, opened 2021-11-25
Maria/bookings crud

### LINKED ISSUE #8: View my bookings' history (M)

### LINKED ISSUE #9: View one booking (M)
Retrieve the details for a booking.

mariatirlea:
Implemented functionalities:
- made a separated class for the bookings controller and implemented the crud operations 
- worked together with @TeoFurcovici on this branch to finish up the update and delete operations that she needed for implementing the end of booking issue

What works: - adding and deleting a booking to/ from database 

What doesn't work : - the get operations because I still get an exception when trying to deserialize the UUID's 

**Update:**
- migrated startDate and endDate type to String because of LocalDateTime deserialization exception 
- the crud methods are all working now 
- the bookingId is autogenerated in constructor 
- I think that if we use String for startDate and endDate we need some kind of validation to ensure that the strings parsed follow a valid LocalDateTime format but  i am not sure yet where should this validation be implemented or if this approach with string dates is ok


src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+            e.printStackTrace();
+            return new ResponseEntity<>(bookings, HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @PostMapping("/delete")
+    public ResponseEntity<Object> deleteBooking(@RequestParam final String id) {
+        try {
+            bookingService.deleteBooking(id);
+            LOGGER.info("DELETE BOOKING - API endpoint invoked");
```
dbernatt: The `LOGGER.info()` should be called before `bookingService.deleteBooking(id)`.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+            e.printStackTrace();
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @PutMapping("/update")
+    public ResponseEntity<Object> updateBooking(@RequestParam final String id, @RequestParam("start") String endDate, PaymentStatus status) {
+        try {
+            bookingService.updateBooking(id, endDate, status);
+            LOGGER.info("UPDATE BOOKING - API endpoint invoked");
```
dbernatt, line 98: Same here. The log should be called before the execution of the service call.

dbernatt (changes requested):

src/main/java/com/bootcamp/demo/model/Booking.java
```diff
@@ -0,0 +1,98 @@
+package com.bootcamp.demo.model;
+
+import java.util.Objects;
+import java.util.UUID;
+
+public final class Booking {
+    private String id;
+    private String serialNumber;
+    private long accountId;
+    private String startDate;
```
dbernatt, line 10: Check this out. I think this resolves the problem with dates. [link](https://www.baeldung.com/spring-date-parameters)

src/main/java/com/bootcamp/demo/model/Booking.java
```diff
@@ -0,0 +1,98 @@
...
+    private String endDate;
+    private PaymentStatus payment;
```
tonks-lupin, line 12: Is this a status of the payment or a payment?

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
tonks-lupin: These types of logs are redundant.

src/main/java/com/bootcamp/demo/model/Booking.java
tonks-lupin: You could as well use long datatype and store the value as millis and convert it in the application via something like:
`Instant.ofEpochSecond(value).atZone(ZoneId.systemDefault()).toLocalDateTime();`
There is also the possibility to create a custom document mapper which would map from QueryDocumentSnapshot to an internal model class.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public ResponseEntity<Object> createBooking(@RequestBody final Booking booking) {
+        try {
+            LOGGER.info("Booking created successfully. Update time: " + bookingService.createBooking(booking));
+            return SUCCESS_RESPONSE;
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.error(e.getMessage());
```
tonks-lupin: You could enhance this log to also contain relevant information regarding when the exception occurred, besides the message itself.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+            e.printStackTrace();
```
tonks-lupin: Logging is sufficient, these could be cleaned up.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+    public ResponseEntity<Object> getBooking(@PathVariable(value = "id") final String id) {
+        Booking booking = null;
+        try {
+            booking = bookingService.getBookingByID(id);
+            LOGGER.info("GET BOOKING BY ID - API endpoint invoked");
+            return new ResponseEntity<>(booking, HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.error(e.getMessage());
+            e.printStackTrace();
+            return new ResponseEntity<>(booking, HttpStatus.BAD_REQUEST);
```
tonks-lupin, line 59: booking will be null in this scenario.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+    @PostMapping("/delete")
+    public ResponseEntity<Object> deleteBooking(@RequestParam final String id) {
+        try {
+            bookingService.deleteBooking(id);
+            LOGGER.info("DELETE BOOKING - API endpoint invoked");
+            return new ResponseEntity<>(HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.error(e.getMessage());
+            e.printStackTrace();
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
```
tonks-lupin: You could reuse the response constants defined above where applicable.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Retrieves a Booking entity by its user's id.
+     */
+    @Override
+    public LinkedHashSet<Booking> getBookings(final String userId) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (userId == null) throw new IllegalArgumentException();
+        LinkedHashSet<Booking> userBookings;
+        userBookings = db.collection(COLLECTION_PATH)
```
tonks-lupin: You could inline this statement directly in a return, no use for creating intermediary variables. 
This applies to other methods as well.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,113 @@
...
+    public BookingService(Firestore db) {
+        this.db = db;
+    }
+
+    /**
+     * Retrieves a Booking entity by its id.
+     */
+    @Override
+    public Booking getBookingByID(final String bookingId) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (bookingId == null) throw new IllegalArgumentException();
```
tonks-lupin: Who/What is the illegal argument here? You could add a message to the exception.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,113 @@
...
+        DocumentReference documentReference = db.collection(COLLECTION_PATH)
+                .document(bookingId);
+        ApiFuture<DocumentSnapshot> future = documentReference.get();
+        DocumentSnapshot document = future.get();
+        Booking booking;
+        if (document.exists()) {
```
tonks-lupin, line 39: The exactly same code can be rewritten with the usage of a ternary if.
But, if you check the toObject implementation you will see that it handles the null scenario.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,106 @@
...
+        this.bookingService = bookingService;
+    }
+
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public ResponseEntity<Object> createBooking(@RequestBody final Booking booking) {
+        try {
+            LOGGER.info("Booking created successfully. Update time: " + bookingService.createBooking(booking));
+            return SUCCESS_RESPONSE;
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
```
tonks-lupin, line 33: It is not really the responsibility of the controller to handle internal exceptions regarding inner service implementation methods - ExecutionException/InterruptedException thrown by futures.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Override
+    public String updateBooking(String bookingId, String endDate, PaymentStatus newStatus) throws ExecutionException, InterruptedException {
+        if (bookingId == null) throw new IllegalArgumentException();
+        DocumentReference docRef = db.collection(COLLECTION_PATH)
+                .document(bookingId);
+        ApiFuture<WriteResult> collectionApiFuture = docRef
+                .update("endDate", endDate,
+                        "payment", newStatus);
+        return collectionApiFuture.get().getUpdateTime().toString();
```
tonks-lupin: You could inline these and format the code accordingly, applies elsewhere as well:
```
        return docRef
                .update("endDate", endDate, "payment", newStatus)
                .get()
                .getUpdateTime()
                .toString();
```

PR #28, opened 2021-12-03
Scooter: Field validation + Exception handling + DTOs + delete a scooter by id.

### LINKED ISSUE #27: Add validation + DTOs +  exception handling to the Scooter Controller
Implement the validation of the input fields.
Crete the necessary DTOs for the Scooter model.
Handle exceptions in the controller and service layer as well, related to the Scooter model.

dbernatt:
Tasks:
- [x] Field validation
- [x] Exception handling
- [x] DTOs
- [x] Delete scooter by id


dbernatt:
> https://github.com/nymphadora-tonks-team/booking-jb/pull/20#discussion_r761813868
If the spacing does not increase readability, it is redundant.
. 
Fixed.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#discussion_r761815472

> What is the advantage of using com.google.common.base.Objects.equal vs java.util.Objects.equals or is this just a wrong import?

I don't know the advantage. It was just a wrong import. :) Fixed.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822489883
> Should case sensitivity be enforced?

No. I changed to ignore-case.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822491851

> Is the logger used?

Yes. If an Exception occure we log the stackTrace. (Exception.class)

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822491851

> If documentation does not add value, it is redundant. You could instead document the purpose of the class itself.

I think the class name is descriptive enough. Removed these comments from the class.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822499210

> If you look into NoHandlerFoundException, you will see that it is thrown if throwExceptionIfNoHandlerFound property is set to true, otherwise the default app behaviour is to return 404 status code.
Does wrapping this into a BadRequest status add any value?

Ups. I don't think so. I change it to 404 then.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822514399

> Considering http status code is already returned, does this add value?
You could stick with a simpler version for exception handling.

Its helpful because on the client side the status code value as number can be used directly without any parsing and checking.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822541387

> Could you highlight the way to reach an IllegalArgumentException here?

No. I think I did this accidentally. :). Fixed.

dbernatt:
https://github.com/nymphadora-tonks-team/booking-jb/pull/20#pullrequestreview-822659917

> Notes regarding the use of DTOs, **which should be discussed with the team**
- when using DTOs, they should be used both for input & output (since the idea is to provide another model, potentially simplified, to the client/user)
- discuss with colleagues about an approach for validation, since if the only reason for having DTOs is for validation (and are as result a wrapper over the model class), perhaps sth else might be more suitable

I use DTOs for validation and data hiding so the user can't send the a Scooter with a battery status. The other reason is the use of reflection which automatically parses the model with the default constructor and I can't specify which constructor is called in that phase. So If I don't use dto for the scooter then I can't limit the user to send a scooter request(POST) without a battery status.

I am open for any other suggestions. But this is the only solution that I see working right know.

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,108 @@
...
+                    .get();
+
+            if (!scooter.exists()) {
+                LOGGER.warn("FIND SCOOTER BY ID - Scooter does not exist! scooterId = {}", scooterId);
+                throw new ItemNotFoundException(String.format("Scooter does not exist. scooterId = %s", scooterId));
+            }
+
+            return scooter.toObject(Scooter.class);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error("FIND SCOOTER BY ID - service function: scooterId = {}.\n Error message: {}.\n StackTrace: {}", scooterId, e.getMessage(), ExceptionUtils.getStackTrace(e));
```
dbernatt: > Why is conversion to string needed as such?

dbernatt: Oh. I realized that I don't need to use a converter, because the error() method have constructor for this type of log.

PR #29, opened 2021-12-06
Vladmsn user crud

vladmsn:
Added service and controller classes for user CRUD operations:

- [x] add
- [x] delete
- [x] getAll
- [x] getById
- [x] getByEmail
- [x] deleteById


src/main/java/com/bootcamp/demo/restapi/UserController.java
```diff
@@ -0,0 +1,78 @@
...
+        } catch (ItemNotFoundException e) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (ServiceException e) {
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @GetMapping("/getUserByEmail/{email}")
+    @ResponseBody
+    public ResponseEntity<Object> getUserByEmail(@PathVariable(value = "email") final String email) {
```
nymphadora-tonks-team, line 64: Notice the duplication between this method and `getUserById`. In order to avoid it, you could extract this to another method which could look like the below:
```
private ResponseEntity<Object> getBy(final Supplier<User> supplier) { // with better/more intuitive name for the method and the param
    try {
        return new ResponseEntity<>(supplier.get(), HttpStatus.OK);
    } catch (ItemNotFoundException e) {
        return new ResponseEntity<>(HttpStatus.NOT_FOUND);
    } catch (ServiceException e) {
    }
}
```
and would be called as `return getBy(() -> userService.getUserById(id));` from `getUserById` and similarly from `getUserByEmail`.

src/main/java/com/bootcamp/demo/restapi/UserController.java
```diff
@@ -0,0 +1,78 @@
...
+        } catch (ItemNotFoundException e) {
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (ServiceException e) {
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @GetMapping
+    @ResponseBody
+    public ResponseEntity<Set<User>> getScooters() {
```
nymphadora-tonks-team: You should rename this in order to be in sync with the dataType that it uses (e.g. `getUsers` instead of `getScooters`).

src/main/java/com/bootcamp/demo/service/IUserService.java
```diff
@@ -0,0 +1,18 @@
+package com.bootcamp.demo.service;
+
+import com.bootcamp.demo.model.User;
+
+import java.util.Set;
+
+public interface IUserService {
```
nymphadora-tonks-team: It does not seem like this is needed (especially since it is only implemented, but not used for injection). See this thread https://github.com/nymphadora-tonks-team/booking-jb/pull/20#discussion_r757110327 , which highlights more details related to this.
Added also a screenshot, since at least for me, the URL does not seem to work ok. Also, the article referenced is this one: https://dimitr.im/spring-interface.
image

src/main/java/com/bootcamp/demo/service/UserService.java
```diff
@@ -0,0 +1,151 @@
...
+    private static final String COLLECTION_USERS_PATH = "bookings/databases/users";
+    private final Firestore db;
+
+    public UserService(Firestore db) {
+        this.db = db;
+    }
+
+    @Override
+    public User createUser(final User user) {
+        LOGGER.info("CREATE USER - service function invoked: user = {}", user.toString());
```
nymphadora-tonks-team: Similar to the above, see this thread for notes related to logging: https://github.com/nymphadora-tonks-team/booking-jb/pull/20#discussion_r762411991. Screenshot below as well in case it does not work ok. URLS:
- https://sematext.com/blog/java-logging-best-practices/
- https://www.geeksforgeeks.org/difference-between-hashig-and-encryption/

image

Also, in this thread https://github.com/nymphadora-tonks-team/booking-jb/pull/20#discussion_r762435389 you can find suggestions for how logs could be shortened, but still provide enough details
image

src/main/java/com/bootcamp/demo/service/UserService.java
```diff
@@ -0,0 +1,151 @@
...
+
+        try {
+            try {
```
nymphadora-tonks-team: It seems a bit weird to have nested `try`s. Also, perhaps this could be simplified to sth like:
- find user by email (without throwing an exception)
- if exists -> throw exception (or update if you think it would be suitable)
- if not exists -> create the user

Note that `set` can be used both for create & update.

src/main/java/com/bootcamp/demo/service/UserService.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Override
+    public void deleteUserById(String userId) {
+        LOGGER.info("DELETE USER BY ID - service function invoked: user = {}", userId);
+
+        try {
+            DocumentSnapshot user = db.collection(COLLECTION_USERS_PATH)
+                    .document(userId)
+                    .get().get();
```
nymphadora-tonks-team, line 61: You could either extract this to another method (to avoid its duplication here & in `getUserById`), or refactor `getUserById` to not throw an exception for missing item (and just return null) & use that method here.

src/main/java/com/bootcamp/demo/service/UserService.java
```diff
@@ -0,0 +1,151 @@
...
+
+
+        } catch (ExecutionException | InterruptedException e) {
+            throw new ServiceException(e);
+        }
+    }
+
+    @Override
+    public User getUserById(final String userId) {
+        LOGGER.info(String.format("FIND USER BY ID - service function invoked. UserId = %s", userId));
```
nymphadora-tonks-team: No need to use `String.format` for this, as the logger has support for this. The below would be equivalent to this:
```
LOGGER.info("FIND USER BY ID - service function invoked. UserId = {}}", userId)
```

src/main/java/com/bootcamp/demo/service/UserService.java
```diff
@@ -0,0 +1,151 @@
...
+    public Set<User> getUsers() {
+        LOGGER.info("GET USERS - service function invoked");
+
+        try {
+            List<QueryDocumentSnapshot> users = db.collection(COLLECTION_USERS_PATH)
+                    .get()
+                    .get()
+                    .getDocuments();
+
+            return StreamSupport.stream(users.spliterator(), false)
```
nymphadora-tonks-team: You can simplify this by using `.stream()` on the documents' list directly. Also, you would not need the variable and could have:
```
return db.collection(COLLECTION_USERS_PATH)
        .get()
        .get()
        .getDocuments()
        .stream()
        .map(qds -> qds.toObject(User.class))
        .collect(Collectors.toSet());
```

src/main/java/com/bootcamp/demo/service/UserService.java
nymphadora-tonks-team: ^ Do not mark the above as resolved if possible since I have referenced it in another PR & it might not redirect properly afterwards 🙂

src/main/java/com/bootcamp/demo/service/IUserService.java
nymphadora-tonks-team: ^ Do not mark the above as resolved if possible since I have referenced it in another PR & it might not redirect properly afterwards 🙂

PR #31, opened 2021-12-06
Crud end of booking

### LINKED ISSUE #6: Book a scooter - end of booking (M)
booking duration -> total trip price -> mocked payment
compute cost by duration
updateStatus - AVAILABLE
update currentLocation of scooter???

### LINKED ISSUE #8: View my bookings' history (M)

### LINKED ISSUE #9: View one booking (M)

mariatirlea:
- fixed the CRUD operations and resolved the PR changes requested in maria/bookings-crud
- added the functionality for the  "Pick a scooter" and "End of Booking" issues, implemented by @TeoFurcovici 
- left the dates as String because we tried many different options like timestamp, Jackson and Spring annotations, custom serializer/deserializer classes but we still got the same errors


src/main/java/com/bootcamp/demo/FirebaseInitializer.java
```diff
@@ -0,0 +1,28 @@
...
+import static com.google.firebase.FirebaseApp.initializeApp;
+import static java.lang.System.getProperty;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+@Service
+public class FirebaseInitializer {
+
+    @PostConstruct
+    public void init() throws IOException {
+        final var serviceAccount = new ByteArrayInputStream(getProperty("firebaseKey").getBytes(UTF_8));
```
nymphadora-tonks-team: Just as a check, this is due to the version used from mater/other branches, right? Same for `FirebaseConfig`.

src/main/java/com/bootcamp/demo/model/Booking.java
```diff
@@ -0,0 +1,104 @@
...
+    private long accountId;
+    private String startDate;
+    private String endDate;
+    private PaymentStatus payment;
+
+    public Booking() {
+    }
+
+    public Booking(String serialNumber, long accountId, String startDate, String endDate, PaymentStatus paymentStatus) {
+        this.id = UUID.randomUUID().toString().replace("-", "");
```
nymphadora-tonks-team: No need to replace "-" with empty. You could simply use the string value of an uuid.

src/main/java/com/bootcamp/demo/model/Booking.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.util.Objects;
+import java.util.UUID;
+
+public final class Booking {
+    private String id;
+    private String serialNumber;
+    private long accountId;
+    private String startDate;
+    private String endDate;
```
nymphadora-tonks-team, line 12: Since these were saved as String, you could have in this class the methods which convert from String -> LocalDateTime, in order to have these in 1 place & not need to duplicate them. Also, as such, the class would be the one responsible for doing the conversion as it is fit (e.g. using always the same DateTimeFormatter).

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public ResponseEntity<Object> createBooking(@RequestBody final Booking booking) {
+        try {
+            LOGGER.info("Booking created successfully. Update time: " + bookingService.createBooking(booking));
+            return SUCCESS_RESPONSE;
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.info("Unfortunately, booking could not be  created .");
```
nymphadora-tonks-team: As a note, you could also log exception details (either with the stacktrace or the message, if the later is enough), as it will provide more details regarding the cause & help in identifying possible issues.
Also, see this article for details about logging: https://sematext.com/blog/java-logging-best-practices/

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.LinkedHashSet;
+import java.util.Random;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Logger;
```
nymphadora-tonks-team: Would it be better to use the Logger from slf4j for consistency across the project (also has some additional features) ?

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+    public BookingsController(BookingService bookingService, ScooterService scooterService) {
+        this.bookingService = bookingService;
+        this.scooterService = scooterService;
+    }
+
+    @PostMapping
+    @ResponseStatus(value = HttpStatus.CREATED)
+    public ResponseEntity<Object> createBooking(@RequestBody final Booking booking) {
+        try {
+            LOGGER.info("Booking created successfully. Update time: " + bookingService.createBooking(booking));
```
nymphadora-tonks-team: You could simplify this to sth like (provided that slf4j will be used, since I think this one does not have support for placeholders): `LOGGER.info("Booking created successfully. Update time: ", bookingService.createBooking(booking));`.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+            return FAILURE_RESPONSE;
+        }
+    }
+
+    @GetMapping("/getBookingsByUserId/{userId}")
+    @ResponseBody
+    public ResponseEntity<LinkedHashSet<Booking>> getBookingsByUserId(@PathVariable(value = "userId") long userId) {
+        LinkedHashSet<Booking> bookingsByUserId = new LinkedHashSet<>();
+        try {
+            LOGGER.info("GET BOOKINGS BY userID - API endpoint invoked");
```
nymphadora-tonks-team: See this https://github.com/nymphadora-tonks-team/booking-jb/pull/29#discussion_r763325800 for more details regarding logging.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.info("Unfortunately, booking could not be  created .");
+            return FAILURE_RESPONSE;
+        }
+    }
+
+    @GetMapping("/getBookingsByUserId/{userId}")
+    @ResponseBody
+    public ResponseEntity<LinkedHashSet<Booking>> getBookingsByUserId(@PathVariable(value = "userId") long userId) {
+        LinkedHashSet<Booking> bookingsByUserId = new LinkedHashSet<>();
```
nymphadora-tonks-team: No need to defined this variable, as you could simply inline the call & not use a collection in the bad request response (as calling the constructor with httpStatus only is equivalent to using null for the value). Thus, you could simplify to sth like:
```
try {
    LOGGER.info("GET BOOKINGS BY userID - API endpoint invoked");
    return new ResponseEntity<>(bookingService.getBookings(userId), HttpStatus.OK);
} catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
    LOGGER.info("Unfortunately,an error happened while trying to retrive a booking based on the user id.");
}
```

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.info("Unfortunately,an error happened while trying to retrive a booking based on the user id.");
+            return new ResponseEntity<>(bookingsByUserId, HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @GetMapping("/getBooking/{id}")
+    @ResponseBody
+    public ResponseEntity<Object> getBooking(@PathVariable(value = "id") final String id) {
+        Booking booking = new Booking();
```
nymphadora-tonks-team: Similar here to the comment above for not using variable & inlining the call.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+            return new ResponseEntity<>(bookingsByUserId, HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.info("Unfortunately,an error happened while trying to retrive a booking based on the user id.");
+            return new ResponseEntity<>(bookingsByUserId, HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @GetMapping("/getBooking/{id}")
+    @ResponseBody
+    public ResponseEntity<Object> getBooking(@PathVariable(value = "id") final String id) {
```
nymphadora-tonks-team: Wondering if this should return `ResponseEntity<Booking>`.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+            LOGGER.info("Unfortunately,an error happened while trying to get a booking.");
+
+            return new ResponseEntity<>(booking, HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @GetMapping("/showAllBookings")
+    @ResponseBody
+    public ResponseEntity<LinkedHashSet<Booking>> getAllBookings() {
+        LinkedHashSet<Booking> bookings = new LinkedHashSet<>();
```
nymphadora-tonks-team: Similar here to the comment above for not using variable & inlining the call.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+    @ResponseBody
+    public void bookScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
+        bookAScooter(lat, longitude);
+
+    }
+
+    @DeleteMapping("/delete")
+    public ResponseEntity<Object> deleteBooking(@RequestParam final String id) {
+        try {
+            LOGGER.info("DELETE BOOKING - API endpoint invoked");
```
nymphadora-tonks-team: No need to log this, as it does not provide useful details.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+            LOGGER.info("Unfortunately,an error happened while trying to get all the bookings.");
+
+            return new ResponseEntity<>(bookings, HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @PostMapping("/bookScooter")
+    @ResponseBody
+    public void bookScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
+        bookAScooter(lat, longitude);
```
nymphadora-tonks-team: Should the scooter be returned here to see which one was booked? Or perhaps its serial number?

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        final double totalCostComputed = booking.getTotalCost(bookingDuration);
+        bookingService.updateBooking(booking.getId(), endDate, booking.getPayment());
+        scooter.setCurrentLocation(new Location(scooter.getCurrentLocation().getLongitude(), scooter.getCurrentLocation().getLongitude()));
+        scooterService.updateScooter(scooter.getSerialNumber(), scooter.getCurrentLocation(), ScooterStatus.AVAILABLE, scooter.getBattery().getLevel());
+
+        return totalCostComputed;
+    }
+
+    public Scooter bookAScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
```
nymphadora-tonks-team: This method is very long and a bit hard to follow. For such cases, it is best to split this is multiple smaller methods which do 1 thing. Look at the suggestion below and check if it is easier to follow, read and understand. Also, find there also some comments (which should not be added after refactoring, but I thought it might be easier to follow as such, than adding separate comments and having to take both those & the refactoring suggesting into account) - for simplicity I have added those with preceded with ////.

```
//// no need to be initialized every time
private static final Random RANDOM = new Random();
....
public Scooter bookAScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
    List<Scooter> allAvailableScooters = scooterService.getAvailableScooters(new Location(lat, longitude), 10.0);
    return !allAvailableScooters.isEmpty()
            ? bookRandomScooter(allAvailableScooters)
            : null;
}

private Scooter bookRandomScooter(List<Scooter> allAvailableScooters) throws ExecutionException, InterruptedException {
    final Scooter pickedScooter = reserveScooter(allAvailableScooters);
    createBooking(pickedScooter);
    return pickedScooter;
}

private void createBooking(Scooter scooter) throws ExecutionException, InterruptedException {
//// Shouldn't here a new booking be created? Or were you thinking that you would first call createBooking & then this one?
//// In my opinion, it might be more intuitive to create the booking here with attributes sth like
////      serialNumber - the one from the scooter
////      accountId - you provide as input the user's id to this method
////      startDate - would be LocalDateTime.now
////      endDate - null
////      payment - null
    final List<Booking> bookings = bookingService.getAllBookings();
    Booking pickedBooking = bookings.get(getRandom(bookings.size()));
    pickedBooking.setSerialNumber(scooter.getSerialNumber());
//// redundant, should be removed
////            String startDateBooking = pickedBooking.getStartDate();
////            pickedBooking.setStartDate(startDateBooking);
    bookingService.createBooking(pickedBooking);
}

private Scooter reserveScooter(final List<Scooter> allAvailableScooters) throws ExecutionException, InterruptedException {
//// Would be simpler to return from the service a list and get the element by calling this directly on the list
//// than having to convert to array. Similar for bookings.
    final Scooter pickedScooter = allAvailableScooters.get(getRandom(allAvailableScooters.size()));
    pickedScooter.setStatus(ScooterStatus.RESERVED);
//// Wondering if here it would be simpler to send the scooter directly.
//// What is the benefit of sending these params & creating a new scooter? Note that if you would use Scooter as param, 
// you could call `createScooter`, since it does the update as well, as it calls set on firestore
    scooterService.updateScooter(pickedScooter.getSerialNumber(), pickedScooter.getCurrentLocation(), pickedScooter.getStatus(), pickedScooter.getBattery().getLevel());
    return pickedScooter;
}

//// should be renamed to be more intuitive
private int getRandom(final int total) {
    return RANDOM.nextInt(total);
}
```

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+    @PutMapping("/update")
+    public ResponseEntity<Object> updateBooking(@RequestParam final String id, @RequestParam("end")
+            String endDate, PaymentStatus status) {
+        try {
+            LOGGER.info("UPDATE BOOKING - API endpoint invoked");
+
+            Booking booking = bookingService.getBookingByID(id);
+            Scooter scooter = scooterService.findScooterById(booking.getSerialNumber());
+
+            if (status == PaymentStatus.SUCCESS) {
```
nymphadora-tonks-team: Should the response be different based on whether the update was done or not? To be able to identify this also in swagger/client.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    public double endBookingAndUpdate(Booking booking, Scooter scooter, String endDate) throws ExecutionException, InterruptedException {
+        long bookingDuration = 0;
+
+        String startDate = booking.getStartDate();
+        //Date endDate = booking.getEndDate();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+        LocalDateTime dateTimeStart = LocalDateTime.parse(startDate, formatter);
```
nymphadora-tonks-team: As suggested in a comment above, this could be moved to `Booking` class.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    public double endBookingAndUpdate(Booking booking, Scooter scooter, String endDate) throws ExecutionException, InterruptedException {
+        long bookingDuration = 0;
+
+        String startDate = booking.getStartDate();
+        //Date endDate = booking.getEndDate();
```
nymphadora-tonks-team: Remove commented out code.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+        LocalDateTime dateTimeStart = LocalDateTime.parse(startDate, formatter);
+        LocalDateTime dateTimeEnd = LocalDateTime.parse(endDate, formatter);
+        if (dateTimeEnd.isAfter(dateTimeStart)) {
+            bookingDuration = dateTimeEnd.getMinute() - dateTimeStart.getMinute();
+        }
+        final double totalCostComputed = booking.getTotalCost(bookingDuration);
+        bookingService.updateBooking(booking.getId(), endDate, booking.getPayment());
+        scooter.setCurrentLocation(new Location(scooter.getCurrentLocation().getLongitude(), scooter.getCurrentLocation().getLongitude()));
```
nymphadora-tonks-team: In order to change the location, shouldn't this be provided as param to the method? Otherwise, this update is unnecessary since it will not change anything. 
Were you thinking that the flow in the end would be sth like
- _create booking + pick scooter & update booking (or pick scooter & create booking, depending on what will be chosen)_
- update scooter with new location
- _end booking_

OR

- _create booking + pick scooter & update booking (or pick scooter & create booking, depending on what will be chosen)_
- _end booking_ & provide scooter location
?

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+            return new ResponseEntity<>(HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.info("Unfortunately,an error happened when trying to update the booking.");
+
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    public double endBookingAndUpdate(Booking booking, Scooter scooter, String endDate) throws ExecutionException, InterruptedException {
```
nymphadora-tonks-team: Similar to the suggestion for `bookAScooter`, do you think this should be split in smaller methods, to ease readability?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+    public BookingService(Firestore db) {
+        this.db = db;
+    }
+
+    /**
+     * Retrieves a Booking entity by its id.
+     */
+    @Override
+    public Booking getBookingByID(final String bookingId) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (bookingId == null) throw new IllegalArgumentException("Parameter 'bookingId' cannot be null");
```
nymphadora-tonks-team: It is best practice to enclose in parantheses the statements from if, even when only 1. This is due to ensuring that you only add what you want to these & not mistakenly execute another instruction as part of if.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+        DocumentReference documentReference = db.collection(COLLECTION_PATH)
+                .document(bookingId);
+        ApiFuture<DocumentSnapshot> future = documentReference.get();
+        DocumentSnapshot document = future.get();
+        Booking booking;
+        if (document.exists()) {
+            booking = document.toObject(Booking.class);
+            return booking;
+        }
+        return null;
```
nymphadora-tonks-team, line 51: You could simply chain the methods directly, without using intermediate variables (since if the doc does not exists, calling `toObject` on it will return null). Thus, you could simplify to:
```
return db.collection(COLLECTION_PATH)
        .document(bookingId)
        .get()
        .get()
        .toObject(Booking.class);
```

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.Comparator;
+import java.util.LinkedHashSet;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+@Service
+public class BookingService implements IBookingService {
+    private static final Logger LOGGER = Logger.getLogger(BookingService.class.getName());
```
nymphadora-tonks-team: You should remove this if it is not used.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    /**
+     * Retrieves all entities of type Booking.
+     */
+    @Override
+    public LinkedHashSet<Booking> getAllBookings() throws ExecutionException, InterruptedException {
+        LinkedHashSet<Booking> allBookings;
+        return allBookings = db.collection(COLLECTION_PATH)
```
nymphadora-tonks-team: No need to use the variable. You could simply return the call to `db.collection...`

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+                .get()
+                .get()
+                .getDocuments()
+                .stream().map(d -> d.toObject(Booking.class))
+                .collect(Collectors.toSet())
+                .stream().sorted(Comparator.comparing(Booking::getStartDate))
```
nymphadora-tonks-team, line 78: No need to call the intermediate `collect` & then `stream` again. This can be simplified to
```
.stream()
.map(d -> d.toObject(Booking.class))
.sorted(Comparator.comparing(Booking::getStartDate))
```

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    /**
+     * Creates a new document containing the given Booking type param.
+     */
+    @Override
+    public String createBooking(final Booking booking) throws ExecutionException, InterruptedException, IllegalArgumentException {
+        if (booking == null) throw new IllegalArgumentException("Parameter 'booking' cannot be null");
```
nymphadora-tonks-team: Same here, you should add { } to the statements from if.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+                .document(booking.getId())
+                .set(booking)
+                .get()
+                .getUpdateTime()
+                .toString();
+    }
+
+    @Override
+    public String deleteBooking(final String bookingId) throws ExecutionException, InterruptedException {
+        if (bookingId == null) throw new IllegalArgumentException("Parameter 'bookingId' cannot be null");
```
nymphadora-tonks-team: Same here, you should add { } to the statements from if.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,112 @@
...
+                .document(bookingId)
+                .delete()
+                .get()
+                .getUpdateTime()
+                .toString();
+    }
+
+    @Override
+    public String updateBooking(String bookingId, String endDate, PaymentStatus newStatus) throws ExecutionException, InterruptedException {
+        if (bookingId == null) throw new IllegalArgumentException("Parameter 'bookingId' cannot be null");
```
nymphadora-tonks-team: Same here, you should add { } to the statements from if.

src/main/java/com/bootcamp/demo/service/IBookingService.java
```diff
@@ -0,0 +1,21 @@
+package com.bootcamp.demo.service;
+
+import com.bootcamp.demo.model.Booking;
+import com.bootcamp.demo.model.PaymentStatus;
+
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+
+public interface IBookingService {
```
nymphadora-tonks-team: Is this left from the version from a branch? As in master, it was not kept due to the reasons outlined here: https://github.com/nymphadora-tonks-team/booking-jb/pull/29#discussion_r763322528

src/main/java/com/bootcamp/demo/service/ScooterService.java
```diff
@@ -0,0 +1,157 @@
...
+
+            LOGGER.info("Created scooter with id = {} at = {}", scooter, updateTime);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error("Failed to create scooter with id = {}.", scooter, e);
+            throw new ServiceException(e);
+        }
+    }
+
+
+    public String updateScooter(final String scooterId, Location location, ScooterStatus newStatus, Double newBatteryLevel) throws ExecutionException, InterruptedException {
```
nymphadora-tonks-team, line 74: Based on the comment from the BookingController, we could check if this is need or not. If kept, you could chain the calls and result in sth like, as it does not look like the variables are needed:
```
Scooter newScooter = new Scooter(scooterId, location, newBatteryLevel, newStatus);
return db.collection(COLLECTION_SCOOTERS_PATH)
        .document(scooterId)
        .set(newScooter)
        .get()
        .getUpdateTime()
        .toString();
```

src/main/java/com/bootcamp/demo/service/assembler/ScooterAssembler.java
```diff
@@ -0,0 +1,11 @@
+package com.bootcamp.demo.service.assembler;
+
+import com.bootcamp.demo.model.Scooter;
+import com.google.cloud.firestore.QueryDocumentSnapshot;
+
+public class ScooterAssembler {
```
nymphadora-tonks-team: This class no longer exists in master. Take care not to add it, since it is not used and as such, it is not needed.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,108 @@
...
+            return new ResponseEntity<>(bookings, HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException e) {
+            LOGGER.error("Failed to find all bookings. {}", e.getMessage());
+            return new ResponseEntity<>(bookings, HttpStatus.BAD_REQUEST);
+        }
+    }
+
+    @PostMapping("/bookScooter")
+    @ResponseBody
+    public ResponseEntity<Scooter> bookScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
```
loredanaiamnitchi: Here I think we need another parameter for searchRadius which is expressed in m

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+        return db.collection(COLLECTION_PATH)
+                .document(bookingId)
+                .update("endDate", endDate, "paymentStatus", newStatus)
+                .get()
+                .getUpdateTime()
+                .toString();
+    }
+
+    public Scooter bookAScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
```
loredanaiamnitchi: Here I think we need another parameter for searchRadius which is expressed in m and not be hardcoded, so that the user can send this parameter.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+        Random random = new Random();
+        Location userLocation = new Location(lat, longitude);
+        Set<Scooter> allAvailableScooters = scooterService.getAvailableScooters(userLocation, 10.0);
```
loredanaiamnitchi: searchRadius = 10m is too small and very unlikely to find scooters. I recommend starting the search with a minimum radius of 100-200m.

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @GetMapping("/getBooking/{id}")
+    public ResponseEntity<Object> getBooking(@PathVariable(value = "id") final String id) {
+        Booking booking = new Booking();
+        try {
+            booking = bookingService.getBookingByID(id);
+            return new ResponseEntity<>(booking, HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.error("Failed to get bookings by id = " + id, e.getMessage());
+            return new ResponseEntity<>(booking, HttpStatus.BAD_REQUEST);
```
tonks-lupin: Is there value in returning an empty booking instance on failure?

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,108 @@
...
+            return FAILURE_RESPONSE;
+        }
+    }
+
+    @PutMapping("/update")
+    public ResponseEntity<Double> updateBooking(@RequestParam final String id, @RequestParam("end")
+            String endDate, PaymentStatus status) {
+        Double totalCostComputed = 0.0;
+        try {
+            Booking booking = bookingService.getBookingByID(id);
```
tonks-lupin, line 83: What if there is no booking found for that id?

src/main/java/com/bootcamp/demo/restapi/BookingsController.java
```diff
@@ -0,0 +1,108 @@
...
+            return SUCCESS_RESPONSE;
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            LOGGER.error("Failed to delete booking with id = " + id, e.getMessage());
+            return FAILURE_RESPONSE;
+        }
+    }
+
+    @PutMapping("/update")
+    public ResponseEntity<Double> updateBooking(@RequestParam final String id, @RequestParam("end")
+            String endDate, PaymentStatus status) {
```
tonks-lupin: Rather than entering payment status here, you could create a PaymentService to imitate a payment action.
For simplicity you can always return successful payment at 1st.
A easy way to imitate and failed payments would be to create a random between a range, let's say 0-10 exclusive, and for a specific interval of values return success for other failure. This would imitate a failure rate of X %.

src/main/java/com/bootcamp/demo/restapi/ScooterController.java
```diff
@@ -49,9 +53,20 @@ public ScooterController(ScooterService scooterService) {
         }
     }
 
+    @PutMapping("/updateScooter")
+    public ResponseEntity<Object> updateScooter(final String scooterId, final Location location, final ScooterStatus newStatus, final Double newBatteryLevel) {
+        try {
+            scooterService.updateScooter(scooterId, location, newStatus, newBatteryLevel);
+            return new ResponseEntity<>(HttpStatus.OK);
+        } catch (ExecutionException | InterruptedException | IllegalArgumentException e) {
+            e.printStackTrace();
```
tonks-lupin: Should the error be logged similar to the other methods?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public Double endBookingAndUpdate(Booking booking, String endDate) throws ExecutionException, InterruptedException {
+        long bookingDuration = 0;
+        Scooter scooter = scooterService.findScooterById(booking.getSerialNumber());
+        String startDate = booking.getStartDate();
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+        LocalDateTime dateTimeStart = LocalDateTime.parse(startDate, formatter);
+        LocalDateTime dateTimeEnd = LocalDateTime.parse(endDate, formatter);
+        if (dateTimeEnd.isAfter(dateTimeStart)) {
+            bookingDuration = dateTimeEnd.getMinute() - dateTimeStart.getMinute();
```
tonks-lupin, line 152: Will this computation work as expected for all scenarios i.e. dateTimeEnd minute is smaller than dateTimeStart minute?
You could find other methods to substract dates and get the result in minutes (hint: Duration class)

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+        final double totalCostComputed = booking.getTotalCost(bookingDuration);
+        updateBooking(booking.getId(), endDate, booking.getPaymentStatus());
```
tonks-lupin: Is the initial status of the booking propagated?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+        scooter.setCurrentLocation(new Location(scooter.getCurrentLocation().getLongitude(), scooter.getCurrentLocation().getLongitude()));
```
tonks-lupin: Will the result be as expected?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+            pickedBooking.setStartDate(startDateBooking);
+            LOGGER.info("booking:" + pickedBooking.toString());
+            createBooking(pickedBooking);
+
+            return pickedScooter;
+        }
+        return null;
+    }
+
+    public Double endBookingAndUpdate(Booking booking, String endDate) throws ExecutionException, InterruptedException {
```
tonks-lupin: You could return a boolean to signify when there was a payment and updates were performed successfully, since as it now it will always return some value for the cost.

src/main/resources/application.properties
```diff
@@ -2,6 +2,10 @@ spring.application.name=booking-java-bootcamp
 
 springdoc.swagger-ui.enabled=true
 springdoc.swagger-ui.path=/swagger.html
+<<<<<<< HEAD
```
tonks-lupin: This seems like a merge conflict, can you verify and update?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public Scooter bookAScooter(final Double lat, final Double longitude) throws ExecutionException, InterruptedException {
+        Random random = new Random();
+        Location userLocation = new Location(lat, longitude);
+        Set<Scooter> allAvailableScooters = scooterService.getAvailableScooters(userLocation, 10.0);
+        if (allAvailableScooters.size() > 0) {
+            int randomIndexScooter = random.nextInt(allAvailableScooters.size());
+            Scooter[] listOfScooters = allAvailableScooters.toArray(new Scooter[allAvailableScooters.size()]);
+            Scooter pickedScooter = listOfScooters[randomIndexScooter];
```
tonks-lupin: You could obtain same result in Java 8:

```
allAvailableScooters.stream()
                .skip(new Random().nextInt(allAvailableScooters.size()))
                .findFirst()
                .orElse(...)
```
Also if you do not really care which scooter is picked at this point you can just do:
```
allAvailableScooters.stream()
                .findAny()
                .orElse(...)
```

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+            pickedScooter.setStatus(ScooterStatus.RESERVED);
+            scooterService.updateScooter(pickedScooter.getSerialNumber(), pickedScooter.getCurrentLocation(), pickedScooter.getStatus(), pickedScooter.getBattery().getLevel());
```
tonks-lupin: When booking a scooter isn't updating status sufficient?
Besides, currentLocation and battery are "empty" updates in the above code.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+            LOGGER.info("scooter:" + pickedScooter.toString());
```
tonks-lupin: If you need this log, you could make it more relevant.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,166 @@
...
+            LinkedHashSet<Booking> bookings = getAllBookings();
```
tonks-lupin: Shouldn't a new booking be created from scratch?

src/main/java/com/bootcamp/demo/model/BookScooter.java
```diff
@@ -0,0 +1,59 @@
+package com.bootcamp.demo.model;
+
+public class BookScooter {
```
nymphadora-tonks-team: Usually, classes' names are nouns, rather than verbs.

src/main/java/com/bootcamp/demo/model/BookScooter.java
```diff
@@ -0,0 +1,59 @@
...
+    private String id;
```
nymphadora-tonks-team: Could you outline what this id represents?

src/main/java/com/bootcamp/demo/model/Booking.java
nymphadora-tonks-team: When resolving a comment, if the suggestion is not implemented, it is usually good to outline why not, in order to discuss & reach an agreement with the other members of the team regarding what is best.

src/main/java/com/bootcamp/demo/model/component/ScooterStatus.java
```diff
@@ -6,11 +6,7 @@
 
     private final String value;
 
-    ScooterStatus(String value) {
-        this.value = value;
-    }
+    ScooterStatus(String value) { this.value = value; }
```
nymphadora-tonks-team: The implementation of this & `getValue` could remain as they were, on separate lines.

src/main/java/com/bootcamp/demo/model/BookScooter.java
loredanaiamnitchi: I'll change it to BookingScooter, it's ok?

src/main/java/com/bootcamp/demo/model/BookScooter.java
loredanaiamnitchi: bookingId, I'll chage name of this parameter from id to bookingId

src/main/java/com/bootcamp/demo/model/BookScooter.java
```diff
@@ -0,0 +1,59 @@
...
+    private String accountId;
+    private Double latitude;
+    private Double longitude;
```
nymphadora-tonks-team: Can't this be a `Location`?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,170 @@
...
+                .update("endDate", endDate, "paymentStatus", newStatus)
+                .get()
+                .getUpdateTime()
+                .toString();
+    }
+
+    public Scooter bookAScooter(BookScooter bookScooter) throws ExecutionException, InterruptedException {
+        List<Scooter> allAvailableScooters = scooterService.getAvailableScooters(
+                new Location(bookScooter.getLatitude(), bookScooter.getLongitude()), bookScooter.getSearchRadius());
+        return  ! allAvailableScooters.isEmpty()
```
nymphadora-tonks-team: Small formatting inconsistency here: extra spaces.

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private Scooter bookRandomScooter(List<Scooter> allAvailableScooters, BookScooter bookScooter) throws ExecutionException, InterruptedException {
+        final Scooter pickedScooter = reserveScooter(allAvailableScooters);
+        createBookingScooter(pickedScooter, bookScooter);
+        return pickedScooter;
+    }
+
+    private void createBookingScooter(Scooter scooter, BookScooter bookScooter) throws ExecutionException, InterruptedException {
+        Booking booking = new Booking();
+        booking.setId(bookScooter.getId());
```
nymphadora-tonks-team: Why is there an id here? Shouldn't the booking be created now? So as such, shouldn't it be a generated id rather than a provided one?

src/main/java/com/bootcamp/demo/service/BookingService.java
```diff
@@ -0,0 +1,170 @@
...
+        if (bookingId == null) throw new IllegalArgumentException("Parameter 'bookingId' cannot be null");
+        return db.collection(COLLECTION_PATH)
+                .document(bookingId)
+                .update("endDate", endDate, "paymentStatus", newStatus)
+                .get()
+                .getUpdateTime()
+                .toString();
+    }
+
+    public Scooter bookAScooter(BookScooter bookScooter) throws ExecutionException, InterruptedException {
```
nymphadora-tonks-team: Wondering if it's better to return a scooter or a booking here?

src/main/java/com/bootcamp/demo/service/IBookingService.java
nymphadora-tonks-team: As a follow-up since I think most conflicts were resolved. Shouldn't this be removed?

src/main/java/com/bootcamp/demo/model/BookScooter.java
loredanaiamnitchi: Yes, I'll change it

src/main/java/com/bootcamp/demo/service/BookingService.java
loredanaiamnitchi: That's right, my bad

src/main/java/com/bootcamp/demo/model/BookScooter.java
loredanaiamnitchi: in fact, I'm going to delete it because we don't need it

src/main/java/com/bootcamp/demo/service/BookingService.java
loredanaiamnitchi: I think it would be good to return a booking to know that the reservation was made successfully and to see its status

src/main/java/com/bootcamp/demo/service/BookingService.java
TeoFurcovici: Fixed. I used both toMinutes and ofMinutes methods from Duration class.

